PR #15, opened 2017-01-04
Moves favicon files to root of site.

philnash:
This has two benefits:

* The docs will automatically pick up the favicon without any work ([see
discussion](https://github.com/crystal-lang/crystal/pull/3832)).
* Less clutter in the `<head>`


spalladino:
@philnash looks ok, but we shouldn't be removing the `link` tags. I understand that browsers will pick the `.ico` by default if there is no `link` tag for the `.png`, and we should be preferring the `.png` version if possible.

philnash:
Fair enough @spalladino, I will replace the links and point to the new location.

philnash:
Although... Chrome and Safari will pick the .ico in that case anyway [according to this blog post](http://www.jonathantneal.com/blog/understand-the-favicon/) and a quick test I just did.

The recommended approach from the blog post if you want to serve PNG favicons to browsers that support them is to just have a `<link>` for the PNG favicon. This will get ignored by browsers that don't support PNGs (IE10 and below) and they will then load the default `/favicon.ico`.

This way we can have PNGs in Chrome and Safari and ICO for fallbacks.

spalladino:
Hmm @matiasgarciaisaia @bcardiff you know more about browser quirks than I do. According to the blogpost, it seems to be the best option. WDYT?

philnash:
I just learned a lot more about favicons than I expected to...

matiasgarciaisaia:
You can do [as much work as you want](https://github.com/joshbuchea/HEAD#favicons) with favicons.

Crystal's website isn't mobile friendly yet, so it doesn't make that much sense to add the touch icons - for now.

I'm OK with having the .ico on the root + a PNG version with the `<link>` tag. We're revisiting this whenever we rewrite Crystal's site _SPOILER ALERT_ soon.

PR #36, opened 2017-03-24
Fix delaunay canvas width when scrollbar present

Alexendoo:
Caused a horizontal scrollbar to appear for browsers using solid scrollbars

Before

image

After

image

### DIFF
--- a/_assets/js/delaunay.js
+++ b/_assets/js/delaunay.js
@@ -3,7 +3,7 @@ $(function() {
     var canvas = d3.select("#delaunay").append("canvas");
 
     function draw() {
-      var width = window.innerWidth,
+      var width = $(window).width(),
           height = $('.header-section').height() + $('nav').height() + 1,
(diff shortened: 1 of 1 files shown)


matiasgarciaisaia:
Thanks for the PR!

It was failing indeed, and this PR solves the issue.

We're using a [staging environment](http://stg.crystal-lang.org/) for Crystal's website now, so I've cherry-picked this commit into [`unstable`](https://github.com/crystal-lang/crystal-website/tree/unstable) instead of just merging it to `master` - that's why I'm closing the PR, too.

You've also made us realize this is going to happen frequently with the PRs, so we'll change the branch<-->environment relationship soon (make `master` be staging, and a `stable` branch be production).

Anyway, the fix will get to production soon. Thank you very much for your contribution! :)

jhass:
Another option is to change the repo's default branch (and if you rename `unstable` to `develop` it makes more sense too).

PR #41, opened 2017-04-14
Correct case on GitHub

jamierocks:
- s/Github/GitHub
- s/git/Git

And, I've just seen the previous two prs doing exactly this. Rest assured I have been bothered to do an actual search using an IDE ;)


mverzilli:
Thank you @jamierocks!

pmallol:
Ok in version c8873ee

PR #44, opened 2017-05-08
Fixed grammar on home page (inspired on => inspired by)

ndbroadbent:
See: https://forum.wordreference.com/threads/inspired-by-inspired-on-something.1628637/


mverzilli:
Good point, thank you!

PR #46, opened 2017-06-08
[grammar/typo] "a~~n~~ union"

noformnocontent:
as of https://github.com/crystal-lang/crystal/commit/951e3f7341589b8abeade6a61d76e1b2ef605b7e

PR #47, opened 2017-06-08
[grammar/typo] "a~~n~~ union"

noformnocontent:
copy of https://github.com/crystal-lang/crystal-website/pull/46 to unstable, since im not sure which one is gonna be pushed to production earlier

### DIFF
--- a/index.html
+++ b/index.html
@@ -53,7 +53,7 @@ custom_body_classes:
   <div class="row">
     <div class="col s12">
       <h2>Null reference checks</h2>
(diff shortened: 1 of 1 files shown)


mverzilli:
Good catch, thanks! And you're right, we use `unstable` as the main development branch in this repo, so it's better to send PRs here. Thank you again!

RX14:
FYI you can change the target branch of PRs by using the edit button as if you were going to change the title.

Issue #43, opened 2017-04-26
Bountysource link in Sponsors instructions doesn't have a text-decoration


pmallol:
Version c8873ee

1. Navigate to [Sponsors](http://stg.crystal-lang.org/sponsors/)
2. Hover over "Bountysource profile"

Expected: 
"Bountysource profile" should appear underlined as other links inside a `<p>` in the page 

![screen shot 2017-04-26 at 10 47 07 am](https://cloud.githubusercontent.com/assets/9214460/25437736/bf3a4284-2a6d-11e7-9f82-95ebf3abbdd1.png)

Note that Chrome and Safari do not support a `<h3>` inside a `<p>` 
https://github.com/crystal-lang/crystal-website/blob/c8873ee173bb9b9c98ac77effbf4a66003c0b69c/sponsors/index.html#L84

PR #48, opened 2017-06-20
add underline hover state to Bountysource link

dannyshafer:
https://github.com/crystal-lang/crystal-website/issues/43

Add underline hover state to Bountysource link

### DIFF
--- a/_assets/css/_home.scss
+++ b/_assets/css/_home.scss
@@ -254,6 +254,10 @@
   }
 }
 
+.bountysource-link:hover {
+  text-decoration: underline;
+}
+
 @media only screen and (max-width: 992px) {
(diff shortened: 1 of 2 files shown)


mverzilli:
Thank you!

bcardiff:
Thanks @dannyshafer, but I think that the following should be done to honor the current style.

The "contact us" has an underline with border-bottom so it is more distant with respect a normal text-decoration. The rule is matching `p > a`. The bountysource link does not has that border-bottom because those paragraph are not wrapped in a `p`. And then we need a rule `h3 + p` to keep that first paragraph with short distance from the title.

Let us know if you want to take care of that, but feel free to say no. Thanks again!

PR #49, opened 2017-08-09
Add the term mailing list for searchbility

Fryguy:
I was trying to find the mailing list and that term was not on the page, so this adds it in.

Note that this was actually the section title prior to 1bf65b101ac9c7f13b71b839672bc0c365895fb9 , which changed it from "Mailing List" to "General Discussion".  I'm not sure why that change was made, so I could also just change the section title if that's preferable.

### DIFF
--- a/community/index.html
+++ b/community/index.html
@@ -28,7 +28,7 @@ nav_order: 4
 {% include community_row.html
           title="General Discussion"
           divId="discuss"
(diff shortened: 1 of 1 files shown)


matiasgarciaisaia:
Thanks @Fryguy!

I think the idea was to focus on what _you'll find there_ rather than _what tool is it_ - even if "Chat room" may not be that on-line with the idea.

PR #53, opened 2017-09-13
Removing benchmarks from homepage

matiasgarciaisaia:
We don't want to encourage benchmarks that compare Crystal with other
languages other than its past-self, so we're dropping the Compiled
section that heavily depended on that.

We may want to reword it later, but we're dropping it for the moment.


straight-shoota:
It's a shame the the whole section is to disappear. Why not leave the text at least? Is there any replacement planned to show off Crystal's magnificent performance? I mean it doesn't need to be a benchmark...

Everything else on the homepage are just functional features. I think it's important to highlight those aspects as well. The fact that it's en par with Go & Co besides having a great syntax made me interested in Crystal. And I think that's what people unfamiliar with the language are interested about.

mverzilli:
We'll probably reintroduce the section once we have well designed and thought out content to put there. Leaving the text by itself didn't look well and seemed not to provide any relevant information other than "this is compiled and fast".

straight-shoota:
So now it's only implicitly stated that Crystal is compiled and uses LLVM... :/

PR #55, opened 2017-10-15
Always generate full-text rss

whtsky:
Currently the rss output of crystal website only contains the excerpt of posts, which makes it unfriendly for RSS reader users.


matiasgarciaisaia:
Great suggestion!

Thanks!

PR #73, opened 2018-12-02
Link to crystalshards.org instead of .xyz, which is down

oprypin:


f:
Fixed. http://crystalshards.xyz/

bcardiff:
We could link to both maybe instead of getting rid of one or another.

And a notice to give visibilty of maintainers would be helpful prpbably

f:
Also open the issues page, if some problem exists again please open an issue: https://github.com/f/crystalshards/issues

f:
@bcardiff What a nice idea! Everyone is happy then :)

straight-shoota:
There is also https://shards.rocks/ (though currently unavailable as well, ping @faustinoaq).

Maybe it would also be a good idea to join forces in order to ensure one service's availability in the future instead of several services where you pick the one that's not currently down.

faustinoaq:
Hi @straight-shoota , yeah shards.rocks is still unavailable, I need some time to get https://github.com/shardsrocks working again :sweat_smile:

PR #75, opened 2019-04-11
working example

mindplay-dk:
I wanted to try the example in the REPL and found it didn't work. I've corrected it.

(Quick question: why are the files named `.md`? This doesn't appear to be valid Markdown? Previews and diffs on GitHub look completely garbled. Just wondering.)

### DIFF
--- a/_includes/sample_macros.md
+++ b/_includes/sample_macros.md
@@ -1,11 +1,16 @@
 {% highlight crystal %}
-class Object
+class Person
+  def initialize(name : String, age : Int)
+    @name = name
+    @age = 0
+  end
+
   def has_instance_var?(name) : Bool
     {% raw %}{{ @type.instance_vars.map &.name.stringify }}{% endraw %}.includes? name
   end
 end
 
 person = Person.new "John", 30
-person.has_instance_var?("name") #=> true
-person.has_instance_var?("birthday") #=> false
+puts person.has_instance_var?("name") #=> true
+puts person.has_instance_var?("birthday") #=> false
 {% endhighlight %}


bew:
It worked though: https://carc.in/#/r/6ph1

You just need to add a `Person` type, like:
```cr
record Person, name : String, age : Int32
```

bcardiff:
I think it would be more interesting to keep the `has_instance_var` in `Object`. But add a plain declaration of `Person` without using `record`

```crystal
class Object
  def has_instance_var?(name) : Bool
    {{ @type.instance_vars.map &.name.stringify }}.includes? name
  end
end

class Person
  def initialize(@name : String, @age : Int32)
  end
end

person = Person.new "John", 30
puts person.has_instance_var?("name")     #=> true
puts person.has_instance_var?("birthday") #=> false
```

@mindplay-dk also the `Int` needs to be some more specific type like `Int32`.

straight-shoota:
> (Quick question: why are the files named .md? This doesn't appear to be valid Markdown? Previews and diffs on GitHub look completely garbled. Just wondering.)

The content is markdown with liquid template. It just looks weird because the only content of this file is a liquid tag which return the Crystal syntax with HTML markup.

mindplay-dk:
Note that I'm totally new to Crystal, I was just drive-by checking it out.

While I was able to figure out why it didn't work, others probably shouldn't have to?

(On that note, it would be lovely if there was a "run" button on the code samples... I take it you care about adoption, since you bother providing a REPL - it's common for language sites to make code samples more useful by linking to the REPL so people can try and modify the code samples, and probably low-hanging fruit since you did the hard work of building a REPL already?)

bcardiff:
@mindplay-dk thanks for the suggestion. Does the changes and intentions mentioned in https://github.com/crystal-lang/crystal-website/pull/75#issuecomment-482111563 make sense to you as a reader? If so, do you want to apply them in the PR?

mindplay-dk:
@bcardiff polluting `Object` seems unnecessary for the sake of the example? doesn't seem to me like behavior that should be encouraged for most cases anyhow - maybe this sort of pattern is idiomatic to Crystal, since it builds in ideas from Ruby, but it's something I would personally tend to avoid.

as for the `Int` comment, I'm not sure I understand what's wrong with `Int`, so maybe you better address this issue yourself? as said, I'm just checking out Crystal and don't know much about anything yet :-)

straight-shoota:
Resolved by #614

PR #83, opened 2019-09-06
Fixes #14 - Agregar OC como opción de sponsorship

Mscebba:


index.html
```diff
@@ -8,10 +8,16 @@
 ---
 <div class="banner center bountysource no-select">
```
straight-shoota: Perhaps rename class `.bountysource` to more general `.sponsorship`?

_assets/css/_home.scss
```diff
@@ -180,12 +180,27 @@
 
   .banner.bountysource {
     margin-top: 0;
-    .bountysource-logo {
-      height: 25px;
+    .flag-logos {
+      display: inline;
```
straight-shoota: I'd prefer `display: flex`, it's easier to handle.

bcardiff: We are fine with inline. As long as it does the jobs. The scope is pretty narrow of it.

PR #84, opened 2019-09-08
[CSS] Fix overlap between .sponsorships and .main-actions

straight-shoota:
Currently, on screen sizes smaller than 992px, `.sponsorship` overlaps `.main-actions` buttons.

image

This PR fixes that.

* removes height calculation from `.header-section`, instead fits `#delaunay` to parent height
* adds a bottom padding to `.header-section #content`
* moves `.main-actions` upward using `transform`

### DIFF
--- a/_assets/css/_generic.scss
+++ b/_assets/css/_generic.scss
@@ -150,6 +150,13 @@ table.no-underline a {
   min-height: 465px;
   width: 100%;
 
+  // This padding makes room for .banner.sponsorship to overlap into .header-section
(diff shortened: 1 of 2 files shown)


bcardiff:
After this PR the delaunay is aligned to bottom instead to the top. It overlaps more with the menu and it goes below the sponsorship area. The delaunay was better before. If you prefer we can pick it from here, unless you want to do more html and css.

straight-shoota:
Oh, I assumed that was intentional. Fixed.

Issue #57, opened 2017-12-21
Refer to the crystal book as "language reference"


mverzilli:

straight-shoota:
Related: crystal-lang/crystal-book#173

bcardiff:
We should create a /doc page to link to /reference and /api with a bit of clarification. Once ready, we should remove the temporal redirection of https://github.com/crystal-lang/crystal-book/issues/173#issuecomment-451552310

PR #86, opened 2019-10-15
Installation instructions: rework for all platforms

ftarulla:


bcardiff:
@ftarulla I think I missed that Kubuntu should cover the same content as ubuntu. Official repo and linuxbrew options should be added aswell and it should be reordered between ubuntu and debian I guess. I'm sorry I didn't see that before.

straight-shoota:
The paths don't work. Links are generated for `/install/<platform>` but the files are at `/install/<platform>.html`. I'm not sure if S3 can be configured to serve these files at a URL without the extensions.

Otherwise, the new pages need a permalink configuration (can be configured centrally in `_config.yml`).

bcardiff:
Yikes. How I missed that in stg. Thanks.

straight-shoota:
A few other notes regarding this change. I didn't get to review this properly before merging.

* I'd reorder the platforms, put the most important ones first. Probably something like macOS, ubuntu, arch, alpine, wsl at the top. Sources and tar.gz should be last. And we should add Docker instructions. Bigger icons would also be nice.
* It would be nice to have `Edit this page` links, directly linking to the edit page here on Github. Might be put under `Found any issues?`.
* We need a `/docs` page that links to installation, API and language reference (cf #57)
* We need a proper `Getting started` guide with a short introduction to Crystal, which can be linked at "Done installing?" "Learn" sounds not very welcoming. Calling it "Getting started" is better.

bcardiff:
Regarding the ordering, the criteria was to put first the distros/platforms with official distribution channels.

As long as the top 4 are the current ones: sources, targz, mac, ubuntu I am happy to accept other criteria.

bcardiff:
Platform detection and ordering based on global user base could be another option.

I wanted to have a clear menu for the users to pick their distro context and discover available options from there. And prevent them to go through ubuntu + snap + linuxbrew for example.

PR #87, opened 2019-10-18
Fixed page not found error

girng:
https://crystal-lang.org/install/on_wsl/on_debian
https://crystal-lang.org/install/on_wsl/on_ubuntu

to

https://crystal-lang.org/install/on_debian
https://crystal-lang.org/install/on_ubuntu

@oprypin will be proud

I think this is how you fix it? I'm used to just vanilla HTML ahrefs. No idea what permalink does :O

### DIFF
--- a/install/on_wsl.md
+++ b/install/on_wsl.md
@@ -2,7 +2,7 @@
 layout: install
 subtitle: On Windows Subsystem for Linux
 exclude: true
-permalink: /install/on_wsl/
+permalink: /install/
 ---
 
 The Crystal compiler doesn't run on Windows _yet_. But Crystal can be used with the [Windows Subsystem for Linux](https://msdn.microsoft.com/en-us/commandline/wsl/about), a compatibility-layer for Linux executables running natively on Windows 10.


bew:
> No idea what permalink does

:scream: 

It's important to only do changes that you actually **understand** and that you can **explain**, otherwise **you have no idea** what is going to happen..
Will it fix the problem? Will it create another bug for later?

Did you test your change? Does it make sense?

girng:
If 2 href links cannot be changed without testing an debugging a site all day.. something deeper is terribly wrong.

bew:
As a general rule, it's not because the change seems trivial and dead simple that it should not be tested, nobody is immune to typos or thinking mistake.

In this case, the lines on the top of the markdown document should describe de current file, this file lives at `/install/on_wsl` and permalink is most probably the url that is going to be used to refer to this current page (not ther pages references from this page).

So I don't think changing the permalink of the page itself makes any sense..

girng:
This should be as simple as editing a file, and literally fixing the href inside the `a` tag. It takes 10 seconds, then 5 seconds to hit save and upload. Then another second or so for it to be live on the server.

What a colossal waste of time. Good lord. Please delete / remove this

PR #89, opened 2019-10-18
Fix links

straight-shoota:
This changes internal links to use absolute URLs for paths that are not in this repository (`/images`, `/reference`, `/api`) in order to make the links work in testing and staging environments. Links to `/docs/` are also changed to `/reference` as per crystal-lang/crystal-book#173.
I'm also fixing a few external links, where I could get the new location easily. It seems some of the github permalinks didn't work, but 🤷‍♂  The remaining broken links are vanished hosts we don't control (maybe archive.org could help, but I don't feel any of them is particularly important).

* @sdogruyol Images in https://crystal-lang.org/2017/11/03/2017-crystal-survey-2017-results.html are missing. They were hosted on https://cdn.pbrd.co which seems to be no longer available. If you still have the originals, we might upload them somewhere else.
* https://irclog.crystal-lang.org does not respond either. That's in our control. Does this need fixing or should it just be removed from the community page?


bcardiff:
irclog.crystal-lang.org is working now but only with http.
Otherwise we can use directly https://freenode.irclog.whitequark.org since it's a redirection to it.

straight-shoota:
@bcardiff Merge?

bcardiff:
Some of the links to old repo commits are failing still https://travis-ci.org/crystal-lang/crystal-website/builds/604506152

PR #91, opened 2019-10-22
Improve install section

straight-shoota:
* First commit adds a few visual improvements to distro icons overview:
  * slightly larger icons and font size
  * flexbox layout to avoid hardcoded proportions
  * add a hover effect to emphasize links
* Second commit adjusts header levels and phrasing and adds horizontal separators to better structure and distinguish alternative installation methods
* Third commit is minor text edits.

### DIFF
--- a/_assets/css/_install.scss
+++ b/_assets/css/_install.scss
@@ -82,3 +82,8 @@
     }
   }
 }
+
+hr.install-separator {
+  margin-top: 5rem;
+
+}
--- a/_assets/css/_install.scss
+++ b/_assets/css/_install.scss
@@ -27,15 +27,29 @@
   grid-template-columns: repeat(auto-fill, minmax(240px, 1fr));
   grid-gap: 5rem 0;
   > div {
-    text-align: center;
     .distro-icon {
-      height: 40px;
-    }
-    img {
-      width: 30px;
-    }
-    a {
-      color: $black;
+      display: flex;
+      flex-direction: column;
+      justify-content: center;
+      align-items: center;
+      color: #444;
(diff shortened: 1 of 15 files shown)


bcardiff:
* I'm ok with the content & section changes
  * The `<hr>` should be styled lighter within the content
* I'm ok with the changes in the css organization

But the visual changes I don't think should be applied
* I don't think the transitions/scaling looks good,
* I don't think the bold font is a good choice
* Changing the size of the icons should come with other icon files since otherwise it will look wrong in some scales.

All these last visual things were decided by the designer so I don't want to make aesthetics changes without his opinion. And these will need to be staged for the next round he is available for Crystal.

So, if we can keep the current aesthetics the rest of the changes are good

straight-shoota:
The current visuals are not great: Already on a full HD screen the icons are so small they seem lost. The font size of the labels (15px) is even smaller than the caption text (18px). But the focus should be on the distro names. They need to stand out, not shy away.

image

Increasing the font and icon size fixes that.

image

Hover effect is debatable, but I think there should be some kind of visual indicator that the icons are actually links that can be clicked.

The image files are already compromised even at lower resolution. Fedora and Arch look blurry, Gentoo is cropped at the bottom. These should ideally be all SVG icons. Then the resolution doesn't matter at all.
I don't know where these icons originate from, but at least a few can be found at https://icons8.com/icon/set/operating/metro
I could try to track down more SVG variants but it would probably be easier to know from where the designer (?) brought in the original images.

straight-shoota:
Superseeded by #540

PR #92, opened 2019-10-22
Add edit-on-github link

straight-shoota:
This PR adds a link to Github's edit feature from each install page to encourage readers to fix issues when they notice something is wrong.

We could just add it as such or extend the link to other pages as well. Maybe in the footer? But I figured this is a good start anyway.

### DIFF
--- a/_layouts/install.html
+++ b/_layouts/install.html
@@ -19,4 +19,8 @@ exclude: true
   <h2>Found any issues?</h2>
   <a href="https://forum.crystal-lang.org/c/help-support" target="_blank" class="btn btn-large btn-flat">Forum
     support</a>.
+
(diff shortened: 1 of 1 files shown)


bcardiff:
I think it looks crooked

Screen Shot 2019-10-22 at 10 01 24

Either buttons should be of same size or they should be vertically stacked.

Screen Shot 2019-10-22 at 10 01 11

Yet, I think there are way too many calls to action with the same weight.

Screen Shot 2019-10-22 at 10 06 09

Also: s/Github/GitHub/

straight-shoota:
I guess the latter is good, but then it should be more distinguished from the rest of the page.
WDYT about adding it to more/all pages, maybe in the footer?

bcardiff:
We can have the edit link in the footer. I guess the license is not needed in all the pages, we could use that space for the link.

straight-shoota:
Closing. This isn't going to be merged. The feature will hopefully be included in the upcoming website redesign.

PR #94, opened 2019-10-22
Refactor permalinks for install pages to centralized config

straight-shoota:
Just a small improvement to DRY.

This does not change the output.

`exclude: true` could be scrapped as well, but that's entirely obsolete with #93 anyway.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -18,6 +18,14 @@ defaults:
       image: /images/icon.png
       layout: post
       is_post: true
+  -
+    scope:
+      path: "install"
+      type: pages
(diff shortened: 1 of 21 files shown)


bcardiff:
Isn't `install/index.html` rendered with install layout after this PR since it is within the declared scope?

straight-shoota:
Yes, I figured I'd need to change that, but for some reason the output was correct that way. It seems I messed up the comparison example.

Now it is really fixed, diff to `_site` from `unstable` is empty. Also resolved conflicts.

PR #95, opened 2019-10-23
Add docs page

straight-shoota:
This is a draft for an overview page at `/docs` listing several documentation resources, as per https://github.com/crystal-lang/crystal-website/issues/57#issuecomment-451552824

<ins>This is what it looks like</ins> (second iteration)
image

<del>This is what it looks like:</del> (first iteration)

image

/cc crystal-lang/crystal-book#173

### DIFF
--- a/_data/nav.yml
+++ b/_data/nav.yml
@@ -9,11 +9,7 @@ main:
   url: /sponsors/
 - label: Community
   url: /community/
-- label: Media
-  url: /media/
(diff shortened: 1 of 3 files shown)


bcardiff:
We should use something more aligned with https://crystal-lang.org/community/ layout and aesthetics. There there is space for explanation and links. Buttons are for call to action rather than content.

PR #97, opened 2019-10-30
Add Friends of Crystal

straight-shoota:
/cc @sdogruyol


bcardiff:
The information of production usage will be available directly on the website instead of linking to that site. Serdar is already aware of this idea.

PR #107, opened 2020-05-04
Remove Bountysource relevance in website

danimiba:
Take it away from home, and put open collective first in every mention in sponsor and in the Donate button.


bcardiff:
FYI Donations via bountysource are still received and will be shown in the /sponsors section.
But we will advertise OpenCollective for new ones.

PR #109, opened 2020-06-04
Reword security section

waj:
To encourage sending emails instead of contacting via Keybase.

We should reformat our footer to include links to important sections like this one. Like https://www.rust-lang.org/. @jkicillof what do you think?

### DIFF
--- a/community/index.html
+++ b/community/index.html
@@ -8,7 +8,8 @@ description: There are so many ways to get involved in Crystal. Here's a list of
           title="Security Issues"
           divId="security"
           content="Should you find any security-related issue, please <strong>do not share them openly</strong> in the issue tracker. We have a dedicated mailbox where we keep track of them.<br /><br />
-          You can encrypt your message <a href=\"https://keybase.io/crystal\" target=\"_blank\">via Keybase</a>, or standard PGP (our fingerprint is <code>5995 C83C D754 BE44 8164 1929 0961 7FD3 7CC0 6B54</code>)."
+          You can encrypt your message <a href=\"https://keybase.io/encrypt\" target=\"_blank\">via Keybase</a> (set recipient <code>crystal</code>),
+          or with our <a href=\"http://keys.gnupg.net/pks/lookup?op=get&search=0x09617FD37CC06B54\" target=\"_blank\">PGP key</a> (our fingerprint is <code>5995 C83C D754 BE44 8164 1929 0961 7FD3 7CC0 6B54</code>) and send it to:"
           link_text="security@manas.tech"
           url="mailto:security@manas.tech"
           icon="security"


community/index.html
```diff
@@ -8,7 +8,8 @@
           title="Security Issues"
           divId="security"
           content="Should you find any security-related issue, please <strong>do not share them openly</strong> in the issue tracker. We have a dedicated mailbox where we keep track of them.<br /><br />
-          You can encrypt your message <a href=\"https://keybase.io/crystal\" target=\"_blank\">via Keybase</a>, or standard PGP (our fingerprint is <code>5995 C83C D754 BE44 8164 1929 0961 7FD3 7CC0 6B54</code>)."
+          You can encrypt your message <a href=\"https://keybase.io/encrypt\" target=\"_blank\">via Keybase</a> (set recipient <code>crystal</code>),
+          or with our <a href=\"http://keys.gnupg.net/pks/lookup?op=get&search=0x09617FD37CC06B54\" target=\"_blank\">PGP key</a> (our fingerprint is <code>5995 C83C D754 BE44 8164 1929 0961 7FD3 7CC0 6B54</code>) and send it to:"
```
straight-shoota, line 12: keys.gnupg.net has been a source of reliability problems before. It doesn't seem to be stable.
For that reason we replaced it with https://keybase.io/crystal/pgp_keys.asc in the installation instructions. It would probably be best to use the same URL here.

waj: Wasn't that the MIT server (https://pgp.mit.edu/)? It's always really slow to me and fail many times with proxy errors. Maybe we can include both links and even embed the public key within the page like [Rust](https://www.rust-lang.org/policies/security). Keybase started to look less appealing to me since they announced to be acquired by Zoom.

waj: You're right we should use the same link anyway. I'll add both just in case.

straight-shoota: Yeah, having multiple options doesn't hurt 👍 

> Keybase started to look less appealing to me since they announced to be acquired by Zoom.

Haha. Me, too. 😆

community/index.html
```diff
@@ -9,7 +9,9 @@
           divId="security"
           content="Should you find any security-related issue, please <strong>do not share them openly</strong> in the issue tracker. We have a dedicated mailbox where we keep track of them.<br /><br />
           You can encrypt your message <a href=\"https://keybase.io/encrypt\" target=\"_blank\">via Keybase</a> (set recipient <code>crystal</code>),
-          or with our <a href=\"http://keys.gnupg.net/pks/lookup?op=get&search=0x09617FD37CC06B54\" target=\"_blank\">PGP key</a> (our fingerprint is <code>5995 C83C D754 BE44 8164 1929 0961 7FD3 7CC0 6B54</code>) and send it to:"
+          or with our <a href=\"https://keybase.io/crystal/pgp_keys.asc\" target=\"_blank\">PGP key</a> (fingerprint <code>5995 C83C D754 BE44 8164 1929 0961 7FD3 7CC0 6B54</code>)
+          also available on <a href=\"http://keys.gnupg.net/pks/lookup?op=get&search=0x09617FD37CC06B54\" target=\"_blank\">MIT's key server</a>,
```
straight-shoota, line 13: gnupg is MIT key server?

waj: No, is not 🤦‍♂️

PR #113, opened 2020-06-21
Update on_centos.md

bkmgit:
Change YUM to DNF, the current package manager


bcardiff:
I'm not a CentOS user but [the wiki states](https://wiki.centos.org/PackageManagement/Yum) that the default is yum.

Ruby advertises yum for example. https://www.ruby-lang.org/en/documentation/installation/#yum

PR #139, opened 2021-05-01
Update on_opensuse.md

Stig124:
OpenSUSE din't have up to date installation instructions
This PR fixes those issues


install/on_opensuse.md
```diff
@@ -1,25 +1,25 @@
 ---
-subtitle: On OpenSUSE
+subtitle: On OpenSUSE and SUSE Linux Enterprise (SLE)
 ---
 
-On OpenSUSE, Crystal can be installed from the official rpm package using Zypper.
+On OpenSUSE, Crystal can be installed from the official repo using Zypper.
```
straight-shoota, line 5:
```suggestion
On OpenSUSE, Crystal can be installed from the official repository hosted on the [Open Build Service](https://build.opensuse.org) using Zypper.
```

straight-shoota (changes requested):
The generic installer script should also work on openSUSE now. The instructions for then installer should be the same as for the other OBS-based distros.
Could you please add that, too?
I think the more explicit approach should be labeled as "Manual setup" as with other rpm and deb based instructions.

Stig124:
I was thinking about that, but I don't think it is time, let me explain :

Zypper is a vendor-sticky package manager, people which have used `bintray` repo in the past might have difficulty changing repos when update come depending on zypper config. The last step I laid out, in my eyes, is really important because it preemptively mitigate the vendor-stickyness issue

Before adding the script as main install method, I think it is fair to wait some time for a majority of peoples to migrate repo.

straight-shoota (approved):
Okay, sounds good.

straight-shoota (approved):

beta-ziliani (commented):
Do we have a way to test this @straight-shoota ?

straight-shoota:
The installation instructions themselves are impossible to test because they need human interpretation.

A formalized test is in [distribution-scripts](https://github.com/crystal-lang/distribution-scripts/blob/master/packages/test/opensuse.bats), based on the installer script (which essentially applies the instruction in a generic way).

beta-ziliani:
:rofl:  of course I didn't mean to test the instructions, but the lines therein :-)

beta-ziliani (approved):

straight-shoota:
Merci @Stig124

PR #141, opened 2021-05-07
Update installer script and installation instructions

straight-shoota:
Pulls in the updated installer from https://github.com/crystal-lang/distribution-scripts/blob/master/packages/scripts/install.sh and fixes a few details around package install instructions.


beta-ziliani (approved):
passed the sniff test

PR #156, opened 2021-06-18
Update conference_talks.csv

pettimart:
Added Taylor Carpenter & Will Harris' talk, in parallel with Carl Hörberg's. Both were moved after the break, which is now immediately after Isaac Sloan's talk


straight-shoota:
I've pushed the updated data export from the current Airtable status. Diff is now smaller and as far as I understand, reflects the intended result: The talk by Taylor and Will is inserted parallel to Carl's.

grafik

straight-shoota:
Merged as #157

PR #169, opened 2021-07-16
1.1.0 release post

beta-ziliani:


_posts/2021-07-16-1.1.0-released.md
```diff
@@ -0,0 +1,114 @@
...
+```crystal
+t = {1, 'a'}
+t2 = {true, *t} # => {true, 1, 'a'}
+typeof(t2)      # => Tuple(Bool, Int32, Char)
+
+def f1(x : {Bool, *{Int32, Char}}); end
+def f2(x : Tuple(Bool, *{Int32, Char})); end
+```
+
+Note that it also works for specifying types.
```
asterite, line 28: The changelog says:

> Support splat expansions inside tuple **and array literals**

So I think we should mention that, because it's really powerful!

Some things you can do now that you couldn't (easily) do before:

```crystal
r = 'a'..'e'

a = ['<', *r, '>'] # => ['<', 'a', 'b', 'c', 'd', 'e', '>']
```

Note that you can splat any enumerable inside literals, not only tuples!

Another example:

```crystal
a = 1..10
b = 5..15

Set{*a, *b} # => Set{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15}
```

So it actually works for array-like types too!

asterite: Another example:

```crystal
a = [1, 2, 3]
b = [0, *a, *a] # => [0, 1, 2, 3, 1, 2, 3]
```

_posts/2021-07-16-1.1.0-released.md
```diff
@@ -0,0 +1,114 @@
...
+2. The type of a while expression is now the union of all break expressions within the body, plus `Nil` if the condition isn't exactly the `true` literal; the while expression returns `nil` if the condition fails. The type of a break expression is computed similarly to a break inside a block or a return (`Nil` if empty, `Tuple` if multiple values).
+
+```crystal
+x = while true
+  break 1, 2  if true
+  break 3     if true
+  break 4, "" if true
+end       # => {1, 2}
+typeof(x) # => (Int32 | Tuple(Int32, Int32 | String))
+```
```
asterite, line 39: Maybe also show a more "real" code? For example:

```crystal
# First the first non-blank line in STDIN
first_non_blank_line =
  while line = gets
    break line unless line.blank?
  end

puts first_non_blank_line
```

_posts/2021-07-16-1.1.0-released.md
beta-ziliani: I added the examples. Thanks!

_posts/2021-07-16-1.1.0-released.md
beta-ziliani: Nice one!

beta-ziliani:
@asterite let me know if the changes are OK. @straight-shoota can you take a look? thx

_posts/2021-07-16-1.1.0-released.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Networking
+
+1. `OAuth2::Client` supports the customization of its `HTTP::Client` node (see [10452](https://github.com/crystal-lang/crystal/pull/10452) for details).
+
+2. New `Socket::Addrinfo#inspect` method to obtain the address's details ([#10775](https://github.com/crystal-lang/crystal/pull/10775).
+
+### Runtime
+
+The segfault handler is now implemented in Crystal, in practice meaning that the compilation process of Crystal is simpler: it does not require the former external `libcrystal`.
```
straight-shoota, line 114: There should be some additional notes for package maintainers.

> For package maintainers, this means building `libcrystal.a` (which was part of the `make deps` recipe) is obsolete now. The C code for this library has been removed and packaging instructions need to be updated.
> `make deps` is no longer necessary for building arbitrary Crystal programs, it only builds `llvm_ext.a` which is required for linking against LLVM (for example when building the compiler).

straight-shoota (commented):
I'd suggest to add references to the respective PRs for every feature. The release post has just short mentions, but readers should be able to get to the details easily.

I also think we should only mention the highlights, not every other new feature. Things like `Addrinfo#inspect`, JSON support for `UUID` or `BigRational#**` with `UInt` are IMO not very relevant additions.

straight-shoota (approved):

PR #214, opened 2021-12-08
Extract release notes as separate collection

straight-shoota:
This change makes use of [Jekyll's collection feature](https://jekyllrb.com/docs/collections/) to separate release notes as a distinct content type. We basically did this already in treating release notes and blog posts differently in the template. Thus this change has no visible effect. The generated result is identical. It's just an internal refactoring for better house keeping.

Practical benefit: We no longer need to update the `latest_version` config data, it's automatically extracted from the latest entry in the releases collection.


beta-ziliani (approved):
This is awesome, thanks! I hated that hack about the latest release.

PR #219, opened 2021-12-15
Release notes for 1.3.0

straight-shoota:
I'm starting this PR early to improve and untangle the release process. We already know most of the things that are going to be released in 1.3.0. The major changes should already be in master. So we can start writing about them.
There is a similar PR for the changelog: https://github.com/crystal-lang/crystal/pull/11599

The following is a list of the most important changes that I think should be mentioned in the release notes. If you have additional suggestions, please comment. The list is open for extensions.

* [x] Interpreter (@asterite, @straight-shoota) (#229)
* [x] Windows Support (@HertzDevil) (https://github.com/crystal-lang/crystal-website/pull/230)
* [x] Unicode (@straight-shoota) (https://github.com/crystal-lang/crystal-website/pull/228)
* [x] Syscalls (@lbguilherme) (https://github.com/crystal-lang/crystal-website/pull/231)
* [x] Multi-Assign (@HertzDevil) (#227)
* [x] Number Autocast (@asterite, @beta-ziliani) (#225)
* [x] Improved Numerical Support / Lexing Number Literals (@BlobCodes, @HertzDevil) (https://github.com/crystal-lang/crystal-website/pull/232)

I tagged some people who I believe did major contributions for the respective changes in this release and hope they might also contribute a brief description for the release notes. Don't feel obliged to do that. I just think whoever implemented something is also best qualified to tell about it. Anyone else can step in as well.
Descriptions should summarize the change and for whom it's relevant. A code example is often helpful. You can draw inspiration from [previous release notes](https://crystal-lang.org/blog/#release_notes). Please submit via pull request against this branch (https://github.com/crystal-lang/crystal-website/tree/release/1.3.0).

PR #227, opened 2022-01-03
Explaining the changes in multi-assignments

beta-ziliani:
Completes a section of #219

### DIFF
--- a/_releases/2022-01-06-1.3.0-released.md
+++ b/_releases/2022-01-06-1.3.0-released.md
@@ -25,17 +25,26 @@ In this release we included [XXX PRs since the 1.2.2 release](https://github.com
 Multiple assignments got improved in various ways. First, it's possible to take a "splat of elements" in a multi-assignment:
 
 <div class="code_section">{% highlight crystal %}
-n_first, *n_s, n_last = [1, 2, 3, 4, 5]       # n_s is [2, 3, 4]
-*t_s, t_last = {"This", 15, 4, "tuple", true} # t_s is {"This", 15, 4, "tuple"}
+# Splat in multi-assign with array
+first, *rest, last = [1, 2, 3, 4, 5]
+first # => 1
+rest # => [2, 3, 4]
+last # => 5
+
+# Splat in multi-assign with tuple
+*rest, last = {"This", 15, 4, "tuple", true}
+rest # => {"This", 15, 4, "tuple"}
+last # => true
 {% endhighlight %}</div>
 
-Second, it's possible to detect unbalanced multi-assignments:
+Second, there's an optional preview feature to detect unbalanced multi-assignments. It can be enabled with the compiler flag `-Dstrict_multi_assign`.
 
 <div class="code_section">{% highlight crystal %}
-not, that, many = [1, 2]    # Multiple assignment count mismatch (IndexError)
+a, b, c = {1, 2} # Error: index out of bounds for Tuple(Int32, Int32) (2 not in -2..1)
+a, b = {1, 2, 3} # Error: cannot assign Tuple(Int32, Int32, Int32) to 2 targets
 {% endhighlight %}</div>
 
-⚠️ This second improvement is a breaking change; therefore, you need to enable it with the flag `-Dstrict_multi_assign`. We encourage to do so to detect possible errors in your code, and to prepare to 2.0, which likely will include it by default. An additional restriction guarded by that flag is that if the right-hand side has a unique element, it must be of type `Indexable` or `Enumerable`.
+⚠️ This second improvement is a breaking change; therefore it's not enabled by default. We encourage using it to detect possible errors in your code. This setting will likely be the default in 2.0. An additional restriction guarded by that flag is that if the right-hand side has a unique element, it must be of type `Indexable` or `Enumerable`.
 
 ## Number Autocast
 
--- a/_releases/2022-01-06-1.3.0-released.md
+++ b/_releases/2022-01-06-1.3.0-released.md
@@ -22,6 +22,21 @@ In this release we included [XXX PRs since the 1.2.2 release](https://github.com
 
 ## Multi-Assign
 
+Multiple assignments got improved in various ways. First, it's possible to take a "splat of elements" in a multi-assignment:
+
+<div class="code_section">{% highlight crystal %}
+n_first, *n_s, n_last = [1, 2, 3, 4, 5]       # n_s is [2, 3, 4]
+*t_s, t_last = {"This", 15, 4, "tuple", true} # t_s is {"This", 15, 4, "tuple"}
+{% endhighlight %}</div>
+
+Second, it's possible to detect unbalanced multi-assignments:
+
+<div class="code_section">{% highlight crystal %}
+not, that, many = [1, 2]    # Multiple assignment count mismatch (IndexError)
+{% endhighlight %}</div>
+
+⚠️ This second improvement is a breaking change; therefore, you need to enable it with the flag `-Dstrict_multi_assign`. We encourage to do so to detect possible errors in your code, and to prepare to 2.0, which likely will include it by default. An additional restriction guarded by that flag is that if the right-hand side has a unique element, it must be of type `Indexable` or `Enumerable`.
+
 ## Number Autocast
 
 ## Improved Numerical Support / Lexing Number Literals
--- a/_releases/2022-01-06-1.3.0-released.md
+++ b/_releases/2022-01-06-1.3.0-released.md
@@ -30,7 +30,9 @@ first, *rest, last = [1, 2, 3, 4, 5]
 first # => 1
 rest # => [2, 3, 4]
 last # => 5
+{% endhighlight %}</div>
 
+<div class="code_section">{% highlight crystal %}
 # Splat in multi-assign with tuple
 *rest, last = {"This", 15, 4, "tuple", true}
 rest # => {"This", 15, 4, "tuple"}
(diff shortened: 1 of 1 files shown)


_releases/2022-01-06-1.3.0-released.md
```diff
@@ -22,6 +22,21 @@ In this release we included [XXX PRs since the 1.2.2 release](https://github.com
...
+*t_s, t_last = {"This", 15, 4, "tuple", true} # t_s is {"This", 15, 4, "tuple"}
+{% endhighlight %}</div>
+
+Second, it's possible to detect unbalanced multi-assignments:
+
+<div class="code_section">{% highlight crystal %}
+not, that, many = [1, 2]    # Multiple assignment count mismatch (IndexError)
+{% endhighlight %}</div>
+
+⚠️ This second improvement is a breaking change; therefore, you need to enable it with the flag `-Dstrict_multi_assign`. We encourage to do so to detect possible errors in your code, and to prepare to 2.0, which likely will include it by default. An additional restriction guarded by that flag is that if the right-hand side has a unique element, it must be of type `Indexable` or `Enumerable`.
```
straight-shoota, line 38:
```suggestion
⚠️ This second improvement is a breaking change; therefore it's not enabled by default. We encourage using it to detect possible errors in your code. This setting will likely be the default in 2.0. An additional restriction guarded by that flag is that if the right-hand side has a unique element, it must be of type `Indexable` or `Enumerable`.
```

_releases/2022-01-06-1.3.0-released.md
```diff
@@ -22,6 +22,21 @@ In this release we included [XXX PRs since the 1.2.2 release](https://github.com
...
 ## Multi-Assign
 
+Multiple assignments got improved in various ways. First, it's possible to take a "splat of elements" in a multi-assignment:
+
+<div class="code_section">{% highlight crystal %}
+n_first, *n_s, n_last = [1, 2, 3, 4, 5]       # n_s is [2, 3, 4]
+*t_s, t_last = {"This", 15, 4, "tuple", true} # t_s is {"This", 15, 4, "tuple"}
+{% endhighlight %}</div>
+
+Second, it's possible to detect unbalanced multi-assignments:
```
straight-shoota, line 32: I'd mention the opt-in restriction right at the start, to make it clear that the following example does not show the default behaviour:
```suggestion
Second, there's an optional preview feature to detect unbalanced multi-assignments. It can be enabled with the compiler flag `-Dstrict_multi_assign`.
```

_releases/2022-01-06-1.3.0-released.md
```diff
@@ -22,6 +22,21 @@ In this release we included [XXX PRs since the 1.2.2 release](https://github.com
...
+
+<div class="code_section">{% highlight crystal %}
+not, that, many = [1, 2]    # Multiple assignment count mismatch (IndexError)
```
straight-shoota, line 35: This is a bad example because the flag doesn't have much effect at runtime (for typical `Indexable` implementations, that is).
Both enabled and disabled raises `IndexError`. Only difference is the message (`Multiple assignment count mismatch` vs. `Index out of bounds`) and location (user code vs. `Indexable#[]`).
I don't think this part of the change is even *worth* mentioning because if that.

Instead, we should use an example of a compile time error. Borrowed and updated from the PR:
```suggestion
a, b, c = {1, 2} # Error: index out of bounds for Tuple(Int32, Int32) (2 not in -2..1)
a, b = {1, 2, 3} # Error: cannot assign Tuple(Int32, Int32, Int32) to 2 targets
```

straight-shoota: Btw. that first error message looks odd. I don't think that's intended 🤔 
It should be `cannot assign Tuple(Int32, Int32) to 3 targets`.

_releases/2022-01-06-1.3.0-released.md
```diff
@@ -22,6 +22,32 @@ In this release we included [XXX PRs since the 1.2.2 release](https://github.com
...
+
+Second, there's an optional preview feature to detect unbalanced multi-assignments. It can be enabled with the compiler flag `-Dstrict_multi_assign`.
+
+<div class="code_section">{% highlight crystal %}
+a, b, c = {1, 2} # Error: index out of bounds for Tuple(Int32, Int32) (2 not in -2..1)
+a, b = {1, 2, 3} # Error: cannot assign Tuple(Int32, Int32, Int32) to 2 targets
+{% endhighlight %}</div>
+
+⚠️ This second improvement is a breaking change; therefore it's not enabled by default. We encourage using it to detect possible errors in your code. This setting will likely be the default in 2.0. An additional restriction guarded by that flag is that if the right-hand side has a unique element, it must be of type `Indexable` or `Enumerable`.
+
```
oprypin, line 50:
```suggestion

If you actually relied on ignoring the "overflow" items, this syntax will achieve the old behavior:

<div class="code_section">{% highlight crystal %}
a, b, *_ = [1, 2, 3]
# Same as the old: a, b = [1, 2, 3]
{% endhighlight %}</div>
```

_releases/2022-01-06-1.3.0-released.md
```diff
@@ -22,6 +22,32 @@ In this release we included [XXX PRs since the 1.2.2 release](https://github.com
...
+# Splat in multi-assign with tuple
+*rest, last = {"This", 15, 4, "tuple", true}
+rest # => {"This", 15, 4, "tuple"}
+last # => true
+{% endhighlight %}</div>
+
+Second, there's an optional preview feature to detect unbalanced multi-assignments. It can be enabled with the compiler flag `-Dstrict_multi_assign`.
+
+<div class="code_section">{% highlight crystal %}
+a, b, c = {1, 2} # Error: index out of bounds for Tuple(Int32, Int32) (2 not in -2..1)
```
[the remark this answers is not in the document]
straight-shoota: See previous comment: https://github.com/crystal-lang/crystal-website/pull/227#discussion_r777708890

_releases/2022-01-06-1.3.0-released.md
straight-shoota: Good addition. But I think this should be moved up at the end of the first part. And instead of a vague "this syntax", I'd explicitly call the underscore splat target.

_releases/2022-01-06-1.3.0-released.md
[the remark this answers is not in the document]
beta-ziliani: but if this errs in 1.2.2, then we should probably remove this example and leave the second one?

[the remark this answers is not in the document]
straight-shoota: I'd still leave this as reference. "This was already an error before for too many values, now there's this error when there are too many values."

_releases/2022-01-06-1.3.0-released.md
straight-shoota: https://github.com/crystal-lang/crystal/issues/11695

_releases/2022-01-06-1.3.0-released.md
```diff
@@ -22,6 +22,46 @@ In this release we included [XXX PRs since the 1.2.2 release](https://github.com
 
 ## Multi-Assign
 
+Multiple assignments got improved in various ways. First, it's possible to take a "splat of elements" in a multi-assignment ([#10410](https://github.com/crystal-lang/crystal/pull/10410)):
```
straight-shoota, line 25:
```suggestion
Multiple assignments got improved in various ways. First, it's possible to use a splat target in a multi-assignment ([#10410](https://github.com/crystal-lang/crystal/pull/10410)):
```

_releases/2022-01-06-1.3.0-released.md
```diff
@@ -22,6 +22,46 @@ In this release we included [XXX PRs since the 1.2.2 release](https://github.com
...
+# Ignoring the elements in the middle
+first, *_, last = {"This", 15, 4, "tuple", true}
+first # => "This"
+last # => true
+{% endhighlight %}</div>
+
+Second, there's an optional preview feature to detect unbalanced multi-assignments ([#11145](https://github.com/crystal-lang/crystal/pull/11145)). It can be enabled with the compiler flag `-Dstrict_multi_assign`.
+<div class="code_section">{% highlight crystal %}
+a, b, c = {1, 2} # Error: index out of bounds for Tuple(Int32, Int32) (2 not in -2..1)
+a, b = {1, 2, 3} # Error: cannot assign Tuple(Int32, Int32, Int32) to 2 targets
```
straight-shoota, line 53:
```suggestion
a, b = {1, 2, 3} # Error: cannot assign Tuple(Int32, Int32, Int32) to 2 targets
a, b, c = {1, 2} # Error: index out of bounds for Tuple(Int32, Int32) (2 not in -2..1) (This was already an error before)
```

_releases/2022-01-06-1.3.0-released.md
```diff
@@ -22,6 +22,46 @@ In this release we included [XXX PRs since the 1.2.2 release](https://github.com
...
+{% endhighlight %}</div>
+
+Note that the first example fails with a similar error in 1.2.2 (this error is [expected to change](https://github.com/crystal-lang/crystal/issues/11695) in the near future).
+
+To ignore the extra items in strict mode you can use the underscore splat notation:
```
straight-shoota, line 58:
```suggestion
To ignore the extra items in strict mode you can use the underscore splat notation. This can be used to port existing code to work with the `strict_multi_assign` flag.
```

_releases/2022-01-06-1.3.0-released.md
beta-ziliani: What's the point of this swap? It's inconsistent with the following paragraph.

straight-shoota: Eh, the swap is irrelevant 🙈 It's about adding a note that the error for `a, b, c = {1, 2}` is not new.

PR #231, opened 2022-01-04
Explaining syscall in release post 1.3.0

beta-ziliani:


_releases/2022-01-06-1.3.0-released.md
```diff
@@ -20,6 +20,19 @@ In this release we included [XXX PRs since the 1.2.2 release](https://github.com
 
 ## Syscalls
 
+We are adding an experimental API to create native system calls in Linux architectures ([#10777](https://github.com/crystal-lang/crystal/pull/10777)). This is a first step towards supporting Linux's io_uring IO interface to improve performance in kernels that support it. For example, the following code writes `"Hello!"` to `stdout` using the call generated by the macro `Syscall.def_syscall`.
```
straight-shoota, line 23: Syscalls are not specific to Linux and should work for many other platforms eventually. Only Linux is supported for now because that's what we started with (motivated by IO uring).

```suggestion
We added an experimental API to create native system calls ([#10777](https://github.com/crystal-lang/crystal/pull/10777)). Only Linux is supported for now. This is a first step towards supporting Linux's `io_uring` interface to improve IO performance.

For example, the following code writes `"Hello!"` to `stdout` using the call generated by the macro `Syscall.def_syscall`.
```
Splitting the last paragraph because the example has nothing to do with `io_uring`. 

I don't think we need to comment on the internal behaviour of the syscall in the example. That's irrelevant for the demonstration. Maybe just copy the wording from the API docs? https://crystal-lang.org/api/master/Syscall.html#def_syscall%28name%2Creturn_type%2C%2Aargs%29-macro
> To define system calls open a module and use the `def_syscall` macro. Pass in the system call name, the return type and its arguments.

beta-ziliani: > I don't think we need to comment on the internal behaviour of the syscall in the example. That's irrelevant for the demonstration.

I have no idea what this is referring to in the original text 🤔 In any case, I will simply put the link to the docs.

straight-shoota: I was refering to ``writes `"Hello!"` to `stdout` ``. For this demonstration it's irrelevant *what* the syscall does, it just matters *how* to call it.

_releases/2022-01-06-1.3.0-released.md
```diff
@@ -20,7 +20,9 @@ In this release we included [XXX PRs since the 1.2.2 release](https://github.com
 
 ## Syscalls
 
-We are adding an experimental API to create native system calls in Linux architectures ([#10777](https://github.com/crystal-lang/crystal/pull/10777)). This is a first step towards supporting Linux's io_uring IO interface to improve performance in kernels that support it. For example, the following code writes `"Hello!"` to `stdout` using the call generated by the macro `Syscall.def_syscall`.
+We added an experimental API to create native system calls ([#10777](https://github.com/crystal-lang/crystal/pull/10777)). Only Linux is supported for now. This is a first step towards supporting Linux's `io_uring` interface to improve IO performance.
+
+For example, the following code writes `"Hello!"` to `stdout` using the call generated by the macro [`Syscall.def_syscall`](https://crystal-lang.org/api/1.3.0/Syscall.html#def_syscall%28name%2Creturn_type%2C%2Aargs%29-macro).
```
straight-shoota, line 25:
```suggestion
To define system calls open a module and use the [`Syscall.def_syscall`](https://crystal-lang.org/api/1.3.0/Syscall.html#def_syscall%28name%2Creturn_type%2C%2Aargs%29-macro) macro. Pass in the system call name, the return type and its arguments. 
```

beta-ziliani: Further refined:
```suggestion
To define system calls open a module and use the [`Syscall.def_syscall`](https://crystal-lang.org/api/1.3.0/Syscall.html#def_syscall%28name%2Creturn_type%2C%2Aargs%29-macro) macro. As demonstrated in the following example, you need to pass in the system call name, the return type and its arguments.
```

PR #236, opened 2022-01-10
Remove `jekyll-assets`

straight-shoota:
The gem `jekyll-assets` hasn't been updated in a while and is incompatible with newer Jekyll versions (it explicitly targets Jekyll 3).

We don't really need it anyways. It's easy to replace by standard features.

* CSS preprocessing works just as well with Jekyll's asset pipeline
* Image assets didn't have any effect. The only thing this plugin did was adding dimensions to the `` tags (which is mostly unnecessary because the stylesheet sets display sizes for (almost) all images) and useless `alt` attributes.
* JS preprocessing solution might not be as nice, but it still works well enough.

The generated result should be functionally equivalent.


beta-ziliani (approved):
LGTM 👍

PR #238, opened 2022-01-10
Remove `jekyll-timeago`

### LINKED ISSUE #52: Blog post dates are incorrect
The blog post dates seem to be incorrect.

I took this image just now to illustrate:

Example

straight-shoota:
This plugin provided a tiny detail: Relative descriptions of a posts' publication time (such as `today` or `yesterday`).
It's a nice gimmick, but not an essential feature and the effect for a global audience is questionable (yesterday may be today in another part of the globe).
Most significantly, it also imposes a strong constraint for a static site: The description is relative so it would need to be regenerated regularly in order to keep up to date with the progress of time.

So I think it's best to drop this. I don't expect anyone would miss it.

In any case, this could be implemented much better as a progressive client-side enhancement.

Resolves #52

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,6 @@ source 'https://rubygems.org'
 
 gem 'jekyll'
 gem 'jekyll-assets', '~> 2.4.0'
-gem 'jekyll-timeago'
 gem 'jekyll-seo-tag'
 gem 'haml'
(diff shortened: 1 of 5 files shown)


beta-ziliani (approved):
Arguably, "6 January 2022" is also relative to the world side the reader is :-)

But definitively, this should be on the client side and not statically built.

straight-shoota:
Assuming UTC it doesn't matter where you sit.

PR #239, opened 2022-01-10
Remove `haml`

straight-shoota:
I believe this gem is a relict from some time when haml templates might've been used in this repo. It doesn't appear to be used for anything nowadays, so it can be removed.


beta-ziliani (approved):
This one brings me memories :-)

PR #246, opened 2022-01-11
Extract htmlproofer from `Gemfile`

straight-shoota:
htmlproofer is independent of the Jekyll setup, so we should not merge its dependencies


straight-shoota:
CI error shows this won't work in isolation because the Ruby version is to old. This is part of #247 now, which updates Ruby.

PR #251, opened 2022-01-14
Refactor layouts

straight-shoota:
This patch refactors the layouts and header includes of the website for a more organized structure.

The main objective is to remove page-specific behaviour from `_layouts/default.html` and `_includes/header.html`. This configuration should not be hidden in generic layouts/includes and instead be driven from the pages directly.

To achieve this, a couple additional layouts are introduced, which leaves us with the following layout strucuture:

* `skeleton`: Basic HTML skeleton with only the top nav and otherwise blank content.
  * `base`: Base layout with site-specific framing: Top nav and footer
    * `page`: Generic layout for pages, including regular header and page content
    * `post`: Layout for blog posts including post header and body
    * `install`: Layout for pages in the install section
    * `home`: Layout for home page (with special header)

This is an internal change only. The generated result is identical (except for some whitespace changes).

### DIFF
--- a/404.html
+++ b/404.html
@@ -1,5 +1,5 @@
 ---
-layout: default
+layout: page
 title: Page not found
 description: "The requested URL doesn't exist. Meet the Crystal programming language at https://crystal-lang.org/"
 custom_body_classes:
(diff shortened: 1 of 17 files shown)


beta-ziliani (approved):
😍 thanks so much for this!

PR #254, opened 2022-01-27
[CSS] Extract `_variables.scss`

straight-shoota:
Extracts SASS variables into a separate file for separation of concerns. `stylesheet.scss` is only responsible for pulling in file modules.


beta-ziliani (approved):
easy-peasy

PR #255, opened 2022-01-27
[CSS] Remove unnecessary mixins

straight-shoota:
The SASS mixins are completely unnecessary and can be removed completely. Those that where actually used did only provide vendor prefixing for some CSS properties. That's no longer necessary because these properties have wide support by all modern browsers.


beta-ziliani (approved):
🚀

PR #256, opened 2022-01-27
Add page with living style guide

straight-shoota:
This adds a new page at `/style-guide` which contains examples of many common design components on this website. It serves as a living style guide to help ensure consistency of this site's style. It's easier to monitor relevant changes on a single page than having to browse several pages in order to find instances of less common design components.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -8,6 +8,8 @@ future: true
 collections:
   releases:
     output: true
+  style_guide:
+    output: false
 
 defaults:
   -
@@ -65,3 +67,4 @@ kramdown:
     block:
       css_class: code_section
 
+repository: crystal-lang/crystal-website
--- a/_style_guide/elements_team-member.md
+++ b/_style_guide/elements_team-member.md
@@ -1,0 +1,18 @@
+---
+title: Team Member
+type: elements
+---
+<div class="team-member">
+            <img src="/assets/authors/asterite.jpg">
+            <div class="team-member__inner">
+              <span class="member_name">Ary Borenzweig</span>
+              <span class="handles">
+                <span class="github_handle"><a href="https://github.com/asterite" title="GitHub profile"><i class="extra-icons github"></i> asterite</a></span>
+                <span class="forum_handle"><a href="https://forum.crystal-lang.org/u/asterite" title="Crystal Forum profile"><i class="material-icons">forum</i> asterite</a></span>
+              </span>
+
+                <span class="member_role">Co-Creator, Steering Council</span>
+
+
+            </div>
+          </div>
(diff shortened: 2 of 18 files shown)


_config.yml
```diff
@@ -65,3 +67,4 @@ kramdown:
     block:
       css_class: code_section
 
+repository: crystal-lang/crystal-website
```
beta-ziliani, line 70: What is this for? It's unrelated to this PR, right?

_style_guide/typography-code-block.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Code Block
+type: typography
+---
+
+```crystal
```
beta-ziliani, line 6: I suggest to add what the block shows (same for the rest of the code blocks below):

```suggestion
Crystal code:
```crystal
```

_style_guide/elements_team-member.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Team Member
+type: elements
+---
+<div class="team-member">
```
beta-ziliani, line 5: The indentation here is broken.

_style_guide/typography-list-ordered.md
```diff
@@ -0,0 +1,17 @@
+---
```
beta-ziliani, line 1: The meaning are flipped. This is an [ordered list](https://www.markdownguide.org/basic-syntax/#ordered-lists):

1. They
2. Needs
3. Numbers

This is an [unordered list](https://www.markdownguide.org/basic-syntax/#unordered-lists):

 * No
 * Numbers
 * Here

beta-ziliani (commented):
This is AWESOME! 🤩  Left some minor comments.

I wonder if it's possible to also display the `.md` to see where the HTML came from.

_config.yml
straight-shoota: It's used in the `component` include to generate links to the source code.

https://github.com/straight-shoota/crystal-website/blob/b88a3177e52dbb946c01db5033dbdd39f6521457/_includes/style-guide/component.html#L5-L6

straight-shoota:
> I wonder if it's possible to also display the .md to see where the HTML came from.

Yeah, this was my original intention as well. However, it proves to be nontrivial.

The capabilities of Jekyll are a bit limited for this kind of thing. `{% include_relative {{ component.path }} %}` mostly works, but I couldn't get it to remove the frontmatter. And it breaks when the component includes another include (which happens for the *Post Teaser*).
This should be easy to solve with a plugin. I'll look into that, but add it as a separate enhancement.

_style_guide/typography-code-block.md
straight-shoota: I don't think this is necessary. The language should be obvious and is clearly visible in the source code.

beta-ziliani: That's true now that you show the md, smarty-pants ;-)

beta-ziliani (approved):
🚀

PR #261, opened 2022-02-01
Fix `/style-guide` permalink

straight-shoota:
Without this override, the path is https://crystal-lang.org/style-guide.html, but it's supposed to be https://crystal-lang.org/style-guide/


straight-shoota:
Yeah, but this way is IMO more convenient. Having a separate folder with a single index file for every page is a nuisance. These folders are completely unnecessary.

PR #264, opened 2022-02-08
Split `_generic.scss` in typography and site styles

straight-shoota:
This patch extracts generic typography styles into a dedicated file for the sake of separating concerns.

### DIFF
--- a/_sass/_site.scss
+++ b/_sass/_site.scss
@@ -1,103 +1,15 @@
 html, .wrapper {
   height: 100%;
 }
-
 body {
-  background: $white;
-  color: $primary-text;
-  font-family: $sans;
   height: 100%;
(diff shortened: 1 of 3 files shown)


beta-ziliani:
I'm not sure if this is the right separation. or at least, it sounds weird to me to have borders and margins as part of "typography".

straight-shoota:
Borders of typographic elements such as headers or buttons belong to those elements.

beta-ziliani:
IMHO (stressing the H), the concern here is "visuals" and applies to every component

beta-ziliani (approved):
To me `typography` is a bad name; it's not about typography, but about `basic` html components. But if this is a used named for such elements, so be it.

PR #275, opened 2022-02-23
Remove materialize

### LINKED ISSUE #268: Remove materialize css
This website uses [materializecss](https://materializecss.com/) as base style framework. However, the styles from that toolkit are mostly unused in the actual display. The site may be based on material design, but it uses a distinct visual language that has diverged in many places, so that we're mostly using custom styles.
Utilization of styles from materialize is less than 4% on the entire site. We can easily manage without it and thus reduce dependencies and download size. We'll also own more of the styles and they are easier to reason about.

I have already identified the parts that are in use:

* Some typography styles (font definitions, some font sizes and colours)
* Content container
* Columns
* Navigation bar

For the time being, we can extract those pieces from materializecss and keep them around as stand-alone components. Or consider replacing them directly with new solutions.

There is also a big chunk of JavaScript involved with this framework. I don't know what exactly it does, but as far as I can tell, the only useful functionality it provides is the toggle for the sidebar on smaller devices. This can be implemented completely in CSS, however and we should be able to easily ditch the entire JavaScript part of materialize.

straight-shoota:
Resolves #268


beta-ziliani:
I found two issues (Firefox): the quote lost its style:

image

compare with:
image

And the menu doesn't work in the smaller size screen version.

straight-shoota:
Argh, right! The small-screen menu was broken. I remember that now from the time I started working on this PR 🙈 
I wanted to refactor the menu entirely to avoid dragging along any ballast. So I'll do that first.

Issue #276, opened 2022-02-24
Remove `Home` from menu


ftarulla:
We can remove `Home` as an item in the menu and have the `Home` link placed beside the icon.

In this PoC, we can see the name Crystal to the right of the icon.
The icon keeps the same behaviour and at the same time we can click in the name to navigate `Home`.

https://user-images.githubusercontent.com/1175827/155442819-63394acb-d529-4740-a34c-8b424f812c3e.mov

Check the PoC in [PR#277](https://github.com/crystal-lang/crystal-website/pull/277)

straight-shoota:
Closing, this is tracked as part of #284

PR #277, opened 2022-02-24
Remove Home from menu. Add clickable title beside icon to navigate Home

ftarulla:
Related #276 

PoC to show the name Crystal beside the icon as a way to navigate back home.
This is only a PoC and we should check the layout in all pages. For example, in the conference page it's not showing great 😢

Screen Shot 2022-02-23 at 23 04 13


straight-shoota:
On pages with a black header background (i.e. conference page), vertical alignment doesn't fit (as seen in the screenshot). I was however wondering if on these pages we might show only the logo without text. Then we'd have to make the logo clickable, though. Not sure how that works out with the 3D interaction gimmick.

Issue #282, opened 2022-02-24
Remove `Conference` from menu


ftarulla:
A `Conference` is a part of creating a `Community` and `Learning`.

In this PoC we move it to the `Community` page.

<img width="1181" alt="Screen Shot 2022-02-23 at 18 57 30" src="https://user-images.githubusercontent.com/1175827/155453430-61d6d221-7de1-4e37-8a2c-f6853206074f.png">

Also, maybe is a good idea to create a `Learn` page where we can have the link to the `conference` page and a link to the `Crystal Book` (which can now be accessed from the `Learn` button in the main page)

Check the PoC in [PR#283](https://github.com/crystal-lang/crystal-website/pull/283)

straight-shoota:
Perhaps we could also use this opportunity to move the conference page to an archive path (such as `/conference/2021-crystal-1.0-conference`). For now, we can redirect from `/conference`, but as soon as a new conference is coming up, we'll want to have that on the main landing page.

We'll also have to think about where to place a link to that page when it becomes relevant (or other somewhat temporary references). I don't think it should be in the main navigation menu. But still very visible on the header. When we're having a conference, we want people to know about it.

straight-shoota:
Closing, this is tracked as part of https://github.com/crystal-lang/crystal-website/issues/284

PR #283, opened 2022-02-24
Remove Conference from menu. Add Conference section in Community page

ftarulla:
Related #282 

This is only a PoC and further improvement is needed before the PR is ready. For example, we need to:
- Add a good description to the `Conference` section.
- Change the icon (right now we are using the `forum` icon)

### DIFF
--- a/_data/nav.yml
+++ b/_data/nav.yml
@@ -9,8 +9,6 @@ main:
   url: /sponsors/
 - label: Community
   url: /community/
-- label: Conference
-  url: /conference/
 - label: Team
   url: /team/
 - label: Docs
--- a/community/index.html
+++ b/community/index.html
@@ -5,6 +5,16 @@ description: There are so many ways to get involved in Crystal. Here's a list of
 <div class="container">
 
 {% include community_row.html
+          title="Crystal 1.0 Conference"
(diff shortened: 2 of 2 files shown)


straight-shoota:
The section on the community should be generic about *Crystal conferences* not just last year's conference as a specific item. 
There could be a list of links to all previous (and upcoming) conferences.

Besides last year's *Crystal 1.0 Conference*, there was also *Raw Crystal* in 2020. There is no single landing page for that, but those links should probably do: 

* https://www.eventbrite.co.uk/e/raw-crystal-2020-tickets-127439094763
* https://forum.crystal-lang.org/t/raw-crystal-conference-talks-on-youtube/2778
 
----

At this point, we should probably move the *Crystal 1.0 Conference* landing page to a long term location (maybe `/conferences/2021-crystal-1.0`?) to make room for the next one 😄 Keeping a redirect from `/conference` for now, of course.

beta-ziliani:
I'd rather:
 [ ] create a page for the raw crystal conf.
 [ ] move the crystal conf 1.0 to it's own page.
 [ ] create a listing page with the two confs (and future ones).
 [ ] move it to the community page.

Issue #284, opened 2022-02-24
Restructure main navigation menu


ftarulla:
# One issue (or Menu) to rule them all 🧙‍♂️
This issue is an _umbrella issue_, and it can be used as a foreword and index for some changes for the site's menu (and to start the discussion)

They are not _big_ changes but can make a difference when starting to navigate the site for the first time (trying to know/learn about Crystal)

The changes are the result of navigating Crystal's site and some other programming languages' sites (for example Ruby, Rust, Dart, Go) while having in mind the following:
- Easy navigation from `Home` to a page (for example `Blog`) and back to `Home`
- Navigation flow from newcomers' pov (both advanced and beginners)
- Can the `Menu` be simpler? (ie can we remove some item without losing the essence)

Here is the list of issue for some specific changes:

- #276 Remove `Home` from menu
- #278 Remove `GitHub` from menu
- #280 Remove `Forum` from menu
- #282 Remove `Conference` from menu

## Also in this issue ...

... there is a description of Adding `Install`, `Learn` and `Try Online`.  These options are buttons on the `main` page but when navigating to other pages are not accessible any more, making it hard for (not only) newcomers to access this important options.

This new options shouldn't be necessary in the `main` page because we have the 3 buttons so it would be good to have a "dynamic" menu.

This issue would be also a good place to rearrange the items making more newcomers friendly. For example we could sort the menu like this:

Install - Learn - Try online - Docs - Tools - Community - Blog - Team - Sponsors

Read this menu like this:
- The first 3 items `Install`, `Learn` and `Try online` are focused on adopting the language
- The next one `Docs` is focused on mastering the language
- Then we focus on building a community (`Community` and `Blog`)
- At the end we have the `team` behind Crystal and the `sponsors` backing it up 

Also notice a **new item** `Tools` linking to (the external) https://github.com/crystal-lang-tools. This element is more focused on adopting the language and it could be placed before the `Docs` item but then, when in the `main` page, it would be the first item in the menu linking to an external site, meaning that the first element on the menu (in the `main` page) navigates away from the site (and that would be a not so good user experience)

Check the PoC in #285

straight-shoota:
I think it's important to keep the number of navigation items low and focused. I'd rather consider a nested navigation structure than having a main navigation with too many entries. When you compare the main navigation of the mentioned websites of other languages, they usually have 6 or 7 entries. This proposal would end up with 9.

[...]

PR #288, opened 2022-03-11
Remove *Docs* page and redirect to *Crystal Book*

straight-shoota:
The [docs page](https://crystal-lang.org/docs) and the landing page of the [Crystal Book](https://crystal-lang.org/reference/) share the same purpose and have a lot of duplication.

The book has essentially taken over the function of the docs page since the introduction of its landing page in https://github.com/crystal-lang/crystal-book/pull/493. It serves as a central hub for documentation materials for Crystal and incorporates not only resources in the book itself, but also in other locations (such as the website). It's much more advanced than the "old" docs page.

This patch removes the *Docs* page and replaces it with a redirect to the book. 

For the redirect I'm adding the plugin `jekyll-redirect-from` which should be a useful helper for similar tasks as well.

This is essentially a revert of #95 🤷


straight-shoota:
Most of the links from the current docs page are also on the Book landing page with two exceptions (well, actually three but we can ignore that the book doesn't link to itself):

* *Media*: Logos and style guide (https://crystal-lang.org/media)
* *Shards documentation*: Documentation for official shards. [crystal-db](http://crystal-lang.github.io/crystal-db/api/latest/)

I'm not sure a reference to the media page fits well into the Book's landing page. I'd rather see this on the website, but we'll have to incorporate it into the efforts to overhault the navigation system (https://github.com/crystal-lang/crystal-website/issues/284).

The API docs to `crystal-db` are so far a solitary item of shard's documentation. I don't think it makes much sense to link them from the language docs. It's an independent shard (even when it's in `crystal-lang` org) and documentation for it is linked from the shard itself. So I think we can just drop this.

beta-ziliani:
Just FTR there is another link missing: the wiki (there are just some inner links to it).

beta-ziliani (approved):
To be merged after #319

beta-ziliani:
I added the wiki to the new (old) docs page:
https://github.com/crystal-lang/crystal-book/pull/628

PR #305, opened 2022-07-01
Merge sponsors from all sources on name and url

mattrberry:
Looks like this only benefits me, since I've sponsored through both bountysource and opencollective :p

I'd like to touch up the way overrides are handled, but figured it's best to keep this change small.

### DIFF
--- a/scripts/merge.cr
+++ b/scripts/merge.cr
@@ -16,13 +16,12 @@ overrides = Array(Sponsor).new
     sponsors = Array(Sponsor).from_json(file)
     sponsors, overrides = sponsors.partition(&.overrides.nil?) if filename == "others.json"
     sponsors.each do |sponsor|
-      sponsor_hash = sponsor.name.hash ^ (sponsor.url || "").hash # relevant fields to merge on
-      all_sponsors_map[sponsor_hash] = sponsor.merge(all_sponsors_map[sponsor_hash]?)
+      all_sponsors_map[sponsor.id] = sponsor.merge(all_sponsors_map[sponsor.id]?)
     end
   end
 end
 
-all_sponsors = all_sponsors_map.values.flatten # select all sponsors now that they've been merged
+all_sponsors = all_sponsors_map.values # select all sponsors now that they've been merged
 
 overrides.each do |sponsor|
   name_to_override = sponsor.overrides.not_nil!
--- a/scripts/merge.cr
+++ b/scripts/merge.cr
@@ -8,16 +8,21 @@ def level(sponsor : Sponsor)
   LEVELS.find { |amount| amount <= sponsor.last_payment.to_i }.not_nil!
 end
 
-all_sponsors = Array(Sponsor).new
+all_sponsors_map = Hash(UInt64, Sponsor).new
 overrides = Array(Sponsor).new
 
 %w(opencollective.json bountysource.json others.json).each do |filename|
   File.open("#{__DIR__}/../_data/#{filename}") do |file|
     sponsors = Array(Sponsor).from_json(file)
     sponsors, overrides = sponsors.partition(&.overrides.nil?) if filename == "others.json"
-    all_sponsors.concat sponsors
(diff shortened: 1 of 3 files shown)


scripts/sponsors.cr
```diff
@@ -13,6 +13,17 @@ record Sponsor, name : String, url : String?, logo : String?, last_payment : Flo
...
   property overrides : String?
+
+  # Merge this sponsor with the given sponsor. Requires that the name and url match.
+  def merge(other : Sponsor?) : Sponsor
+    return self if other.nil?
+    raise "Can't merge #{self} with #{other}" if name != other.name || url != other.url
+    copy_with(
+      last_payment: since > other.since ? last_payment : other.last_payment,
+      all_time: all_time + other.all_time,
+      since: since < other.since ? since : other.since,
```
straight-shoota, line 24:
```suggestion
      last_payment: Math.max(last_payment, other.last_payment),
      all_time: all_time + other.all_time,
      since: Math.min(since : other.since),
```

scripts/merge.cr
```diff
@@ -8,17 +8,22 @@ def level(sponsor : Sponsor)
...
     sponsors, overrides = sponsors.partition(&.overrides.nil?) if filename == "others.json"
-    all_sponsors.concat sponsors
+    sponsors.each do |sponsor|
+      sponsor_hash = sponsor.name.hash ^ (sponsor.url || "").hash # relevant fields to merge on
+      all_sponsors_map[sponsor_hash] = sponsor.merge(all_sponsors_map[sponsor_hash]?)
+    end
   end
 end
 
+all_sponsors = all_sponsors_map.values.flatten # select all sponsors now that they've been merged
```
straight-shoota, line 25:
```suggestion
all_sponsors = all_sponsors_map.values # select all sponsors now that they've been merged
```
`flatten` shouldn't be necessary because the array is already flat.

scripts/merge.cr
```diff
@@ -8,17 +8,22 @@ def level(sponsor : Sponsor)
...
+all_sponsors_map = Hash(UInt64, Sponsor).new
 overrides = Array(Sponsor).new
 
 %w(opencollective.json bountysource.json others.json).each do |filename|
   File.open("#{__DIR__}/../_data/#{filename}") do |file|
     sponsors = Array(Sponsor).from_json(file)
     sponsors, overrides = sponsors.partition(&.overrides.nil?) if filename == "others.json"
-    all_sponsors.concat sponsors
+    sponsors.each do |sponsor|
+      sponsor_hash = sponsor.name.hash ^ (sponsor.url || "").hash # relevant fields to merge on
```
straight-shoota, line 19: I'd suggest to move the hashing to a method such as `Sponsor#id`.

straight-shoota (approved):
I think this change looks fine, but I have some suggestions for improving the code a little bit.

scripts/sponsors.cr
mattrberry: Happy to add an import to math and change `since`, but changing `last_payment` in that way changes the meaning. For example, if I contributed $100 on bountysource last in 2019, but only $5 on opencollective today, it'd take the bountysource value. Currently, I have it selecting the most recent

straight-shoota: Oops, sorry. Scrap the `last_payment` change.

You don't need to import `Math`, it's already in core lib.

mattrberry:
Apologies for the force push. Just rebased on the latest changes.

mattrberry:
Now that updating sponsors.csv is an automatic change, I can also drop that file from this PR if you prefer.

beta-ziliani:
Thanks @mattrberry and sorry for the delay. Please go ahead and drop the csv change se we can properly merge this.

beta-ziliani (approved):
Awesome, thanks!

mattrberry:
I figure this will pass CI with your latest changes in https://github.com/crystal-lang/crystal-website/pull/311

PR #316, opened 2022-07-28
Adding rel="sponsored nofollow" to sponsors' links

beta-ziliani:
In order to prevent lowering the rank of the site, we mark each sponsor link with `rel="sponsored"`, also adding `nofollow` to those that do not meet a certain criteria. Currently, the criteria is having an icon, that corresponds to the >=75usd tier.

### DIFF
--- a/sponsors/index.html
+++ b/sponsors/index.html
@@ -29,7 +29,7 @@ layout: base
                 {% if sp.logo %}
                   <div class="logo">
                     {% if sp.url %}
-                      <a href="{{sp.url}}" target="_blank" rel="sponsored">
+                      <a href="{{sp.url}}" target="_blank" rel="sponsored nofollow">
                         <img src="/assets/{{ sp.logo }}" />
                       </a>
                     {% else %}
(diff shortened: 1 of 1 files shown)


straight-shoota:
AFAIU, both `sponsored` and `nofollow` have the same effect in marking a link to be ignored. `sponsored` is just a more recent enhancement to describe the nature of the link. It's supported by Google and probably some more search engines.
The point is, combining them doesn't change the effect, and the difference between `rel="sponsored"` and `rel="sponsored nofollow` is only that the latter is correctly understood by agents that don't recognized `sponsored`.

So I would just use `rel="sponsored nofollow` on all links for maximized compatibility.

beta-ziliani:
The point is that if bots follow those that aren't "nofollow", then that's fine.

straight-shoota:
If we wanted links to be followed, then they shouldn't be `rel="sponsored"` either because that just means `nofollow`. It doesn't make any sense to use `sponsored nofollow` on some links and only `sponsored` on others. That just means that agents who don't understand `sponsored` won't be informed about the nature of the link when `nofollow` is missing.

beta-ziliani:
For Google it makes a difference, but I agree that outside Google it might be problematic. Fixed it.

straight-shoota (approved):

PR #323, opened 2022-08-17
adding used in prod page

beta-ziliani:


bcardiff (commented):
The `assets/icons.svg` changes seems wront. The stackoverflow icon is broken.

A link/notice in https://github.com/crystal-lang/crystal/wiki/Used-in-production should be placed so people stop editing that one.

I kind of feel that the layout for mobile devise is not optimal for this kind of list.

beta-ziliani:
Fixed the icon. An edit to the wiki will follow (I'd clean the page, so no outdated info stays there hanging forever).

About the layout, I improved it slightly, the idea is to change it once we have an updated list (and time...).

PR #344, opened 2022-10-06
1.6.0 release post

beta-ziliani:


HertzDevil:
We need a section about deprecating suffix-less `UInt64` literals (that don't fit into an `Int64`)

_releases/2022-10-06-1.6.0-released.md
```diff
@@ -0,0 +1,73 @@
...
+
+By making some console methods available in Windows, a significant [refactor](https://github.com/crystal-lang/crystal/pull/12352) lead to deprecating three macro methods in `FileDescriptor`: `cooked_from_tc_mode!`, `noecho_from_tc_mode!`, and `raw_from_tc_mode!`. Additionally, methods `#noecho!` and `#raw!` now return `nil`.
+
+## ⚠️ Refactors in `File::Info`
+
+[Another refactor](https://github.com/crystal-lang/crystal/pull/12385) lead to an improvement in the `File` API, which was breaking the abstraction by returning an object intended to be internal. In practice, if for some reason you ended up with an object of type `Crystal::System::FileInfo`, it should now be `File::Info`.
+
+## ⚠️ Improvements in overload ordering
+
+[Union types are now considered first](https://github.com/crystal-lang/crystal/pull/12335) when searching for a suitable overload. Consider the following code:
```
HertzDevil, line 30:
```suggestion
[Union types may now be considered before a wider variety of types](https://github.com/crystal-lang/crystal/pull/12335) when searching for a suitable overload. Consider the following code:
```

_releases/2022-10-06-1.6.0-released.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Stats
+
+In this release we included [183 changes since the 1.5.1 release](https://github.com/crystal-lang/crystal/pulls?q=is%3Apr+milestone%3A1.6.0) by 26 contributors. We thank all the effort put into improving the language! ❤️
+
+Below we list the most remarkable changes in the language and stdlib.
+
+## ⚠️ Refactor in console methods
+
+By making some console methods available in Windows, a significant [refactor](https://github.com/crystal-lang/crystal/pull/12352) lead to deprecating three macro methods in `FileDescriptor`: `cooked_from_tc_mode!`, `noecho_from_tc_mode!`, and `raw_from_tc_mode!`. Additionally, methods `#noecho!` and `#raw!` now return `nil`.
```
straight-shoota, line 22: I don't think this is actually a very relevant change for users. 
Might make sense to mention it because of the deprecation, but it should be down at the bottom, not the very first item.

_releases/2022-10-06-1.6.0-released.md
```diff
@@ -0,0 +1,73 @@
...
+
+## ⚠️ Refactors in `File::Info`
+
+[Another refactor](https://github.com/crystal-lang/crystal/pull/12385) lead to an improvement in the `File` API, which was breaking the abstraction by returning an object intended to be internal. In practice, if for some reason you ended up with an object of type `Crystal::System::FileInfo`, it should now be `File::Info`.
```
straight-shoota, line 26: Ditto. This is also a very irrelevant change for users. It should not be on the top.

_releases/2022-10-06-1.6.0-released.md
```diff
@@ -0,0 +1,73 @@
...
+
+Several other [improvements](https://github.com/crystal-lang/crystal/pull/10711) where added to the overloading search algorithm, but only under a special flag `-Dpreview_overload_order`.
+
+## Improvements in the interpreter
+
+In this release [a number interpreter bugs](https://github.com/crystal-lang/crystal/pulls?q=is%3Apr+sort%3Aupdated-desc+milestone%3A1.6.0+label%3Atopic%3Acompiler%3Ainterpreter) got fixed. We are still working towards incorporating it in the major platforms; in the meantime you can build the compiler with interpreter support locally with `make interpreter=1`, see the [interpreter introduction post for more info](https://crystal-lang.org/2021/12/29/crystal-i.html).
+
+## Performance improvements
+
+[Little improvements](https://github.com/crystal-lang/crystal/pulls?q=is%3Apr+sort%3Aupdated-desc+milestone%3A1.6.0+label%3Aperformance) in the compiler and the stdlib went into this release. We don't have a benchmark for you, but users reported non-negligible improvements in compilation speed.
```
straight-shoota: It's worth mentioning that the improvements are not just about speed but also memory consumption.

Also can we do a quick benchmark to say something like "hello world program compiles 7% faster and uses 10% less memory with 1.6.0"?

_releases/2022-10-06-1.6.0-released.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Improvements in the compiler for Windows (native)
+
+[Several improvements](https://github.com/crystal-lang/crystal/pulls?q=is%3Apr+sort%3Aupdated-desc+milestone%3A1.6.0+label%3Aplatform%3Awindows) landed in this release regarding native Windows support. Most notably, [support for building the interpreter](https://github.com/crystal-lang/crystal/pull/12397), [support for libffi](https://github.com/crystal-lang/crystal/pull/12200), and [mutex support](https://github.com/crystal-lang/crystal/pull/12213).
```
straight-shoota: What's non-native Windows support and why is it worth pointing that out?
```suggestion
## Improvements in the compiler for Windows

[Several improvements](https://github.com/crystal-lang/crystal/pulls?q=is%3Apr+sort%3Aupdated-desc+milestone%3A1.6.0+label%3Aplatform%3Awindows) landed in this release regarding Windows support. Most notably, [support for building the interpreter](https://github.com/crystal-lang/crystal/pull/12397), [support for libffi](https://github.com/crystal-lang/crystal/pull/12200), and [mutex support](https://github.com/crystal-lang/crystal/pull/12213).
```

_releases/2022-10-06-1.6.0-released.md
straight-shoota: The increased IO buffer size might worth an explicit mention (https://github.com/crystal-lang/crystal/pull/12507).

_releases/2022-10-06-1.6.0-released.md
```diff
@@ -0,0 +1,73 @@
...
+def foo(x : Bar1 | Bar2)
+  true
+end
+
+foo(Bar1.new)
+```
+
+Before 1.6.0 it would print `'a'`, meaning the generic case was considered first. Now it correctly prints `true`.
+
+Several other [improvements](https://github.com/crystal-lang/crystal/pull/10711) where added to the overloading search algorithm, but only under a special flag `-Dpreview_overload_order`.
```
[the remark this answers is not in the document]
beta-ziliani: My thoughts exactly

_releases/2022-10-06-1.6.0-released.md
```diff
@@ -0,0 +1,73 @@
...
+end
+
+def foo(x : Bar1 | Bar2)
+  true
+end
+
+foo(Bar1.new)
+```
+
+Before 1.6.0 it would print `'a'`, meaning the generic case was considered first. Now it correctly prints `true`.
```
straight-shoota, line 55: We should clearly point out that this bug fix involves a hard breaking change that can lead to programs not building any more. And give advise on what to do when you encounter such a problem (probably restructure the overload type restrictions?). I suppose we didn't put a flag to disable this? @HertzDevil

straight-shoota (commented):
Other changes that might be relevant:

* `Dir.pwd` respect `$PWD`: https://github.com/crystal-lang/crystal/pull/12471
* Improvements in WASM support (?)
* Unicode update: https://github.com/crystal-lang/crystal/pull/12479
* Support for unicode normalization: https://github.com/crystal-lang/crystal/pull/11226

_releases/2022-10-06-1.6.0-released.md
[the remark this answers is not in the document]
straight-shoota: A separate blog post seems like a great idea. But then we don't need to mention it here at all.
I would suggest to remove this sentence from the release post. It's not relevant information. We don't want encourage people to use it without further explanations.

_releases/2022-10-06-1.6.0-released.md
HertzDevil: A migration guide would be something like:

* This happens whenever a parameter has a `Union` restriction in one overload, and the same parameter in another overload has a restriction that isn't a `Path` or an `Underscore`.
* To preserve the old ordering, remove the union overload;
* To ensure the new ordering applies to older releases, move the union overload above the generic one. (This is because previous Crystal versions consider the two overloads unordered.)

_releases/2022-10-06-1.6.0-released.md
beta-ziliani: Actually, given that we found a shard that broke because of the one below, let me keep them on top.

_releases/2022-10-06-1.6.0-released.md
beta-ziliani: What do you mean by "remove the union overload"? Remove the type restriction with the union?

HertzDevil: No, just straight up deleting the `foo(x : Bar1 | Bar2)` overload because it never matches in existing versions.

beta-ziliani:
> Other changes that might be relevant:
> 
>     * `Dir.pwd` respect `$PWD`: [crystal-lang/crystal#12471](https://github.com/crystal-lang/crystal/pull/12471)
> 
>     * Improvements in WASM support (?)
> 
>     * Unicode update: [crystal-lang/crystal#12479](https://github.com/crystal-lang/crystal/pull/12479)
> 
>     * Support for unicode normalization: [crystal-lang/crystal#11226](https://github.com/crystal-lang/crystal/pull/11226)

I added a small addendum with these, except for the WASM part, since it's only a couple of PRs. I prefer to wait to have something more substantial to report.

_releases/2022-10-06-1.6.0-released.md
straight-shoota: Sadly, I couldn't actually verify any noticeable improvement between 1.5.1 and 1.6.0. If anything changed, it's worsening 😢 
But that is probably attributed to other changes, of course (not sure what exactly, though). If at all, these performance improvements reduce the impact.
We need an benchmark suite to quantify the performance effects of compiler changes (https://github.com/crystal-lang/crystal/issues/5508).

_releases/2022-10-06-1.6.0-released.md
```diff
@@ -0,0 +1,85 @@
...
+By making some console methods available in Windows, a significant [refactor](https://github.com/crystal-lang/crystal/pull/12352) lead to deprecating three macro methods in `FileDescriptor`: `cooked_from_tc_mode!`, `noecho_from_tc_mode!`, and `raw_from_tc_mode!`. Additionally, methods `#noecho!` and `#raw!` now return `nil`.
+
+## ⚠️ Refactors in `File::Info`
+
+[Another refactor](https://github.com/crystal-lang/crystal/pull/12385) lead to an improvement in the `File` API, which was breaking the abstraction by returning an object intended to be internal. In practice, if for some reason you ended up with an object of type `Crystal::System::FileInfo`, it should now be `File::Info`.
+
+## Improvements in the interpreter
+
+In this release [a number interpreter bugs](https://github.com/crystal-lang/crystal/pulls?q=is%3Apr+sort%3Aupdated-desc+milestone%3A1.6.0+label%3Atopic%3Acompiler%3Ainterpreter) got fixed. We are still working towards incorporating it in the major platforms; in the meantime you can build the compiler with interpreter support locally with `make interpreter=1`, see the [interpreter introduction post for more info](https://crystal-lang.org/2021/12/29/crystal-i.html).
```
HertzDevil, line 67:
```suggestion
In this release [a number of interpreter bugs](https://github.com/crystal-lang/crystal/pulls?q=is%3Apr+sort%3Aupdated-desc+milestone%3A1.6.0+label%3Atopic%3Acompiler%3Ainterpreter) got fixed. We are still working towards incorporating it in the major platforms; in the meantime you can build the compiler with interpreter support locally with `make interpreter=1`, see the [interpreter introduction post for more info](https://crystal-lang.org/2021/12/29/crystal-i.html).
```

HertzDevil (approved):

straight-shoota (approved):

beta-ziliani:
Why Quinton's approval isn't green? He's a crystaller 🤔

Issue #346, opened 2022-10-10
Pretty URLs: Problems with `.html` extension in file names


straight-shoota:
In Jekyll, the output paths for pages have a `.html` extension by default. For example: `/community/governance.html`
This is not very pretty and practically irrelevant. 
Users typically expect extension-free URLs. An example for that are the links to https://crystal-lang.org/community/governance from crystal-book. This URL is currently a 404 because the extension is missing. It must be https://crystal-lang.org/community/governance.html

A trivial fix for the correct URL is in https://github.com/crystal-lang/crystal-book/pull/643

But it's a general problem that people write pretty URLs. You shouldn't have to type `.html` at the end of a URL.

I propose to change the default url format for pages to `permalink: /:path/:basename/`. This generates pretty urls (it basically creates folders with a `index.html` which the web server then serves for the folder path).

A superior solution to keep both variants (for not breaking existing links) would be to make the `.html` extension optional directly in the web server. For nginx config this would be something like `try_files $uri $uri.html $uri/ =404;`.
I presume there is nothing like this for S3 where we're hosting the website, right? @matiasgarciaisaia any idea?

matiasgarciaisaia:
We're hosting on AWS S3, which supports [object-level redirections](https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-page-redirect.html#redirect-requests-object-metadata).

It'd be awesome if we found a Jekyll plugin that takes care of setting up the redirections, but, if not, we could manually list every URL we currently have (ie, the current sitemap), then enable pretty URLs, deploy and set up each redirection we need.

The general redirection rules match by prefix (not suffix), so we won't be able to use a single, general rule.

straight-shoota:
Yeah, I don't think custom rules in S3 for every single path are a practical solution.

So probably a good path forward is to setup Jekyll to generate pretty URLs by default. For existing pages with `.html` paths, we can create custom redirects to the pretty paths inside Jekyll.

I would ignore blog posts for now.

That leaves us with only a few paths:
```console
$ find _site -name '*.html' -not -name index.html -not -path '_site/20*'
_site/community/governance.html
_site/docs.html
_site/sponsors/original-sponsors.html
_site/learning/crystal_programming.html
_site/404.html
```

`404.html` is only used internally, so we can ignore that.

matiasgarciaisaia:
Is there any reason to avoid the blogposts? I'd do it for every page, so we forget this existed 😇

[...]

PR #352, opened 2022-10-13
Enable pretty URLs for pages

straight-shoota:
This changes the paths for pages to use pretty URLs.

It currently affects the following paths:
```
community/governance.html -> community/governance
docs.html -> docs
sponsors/original-sponsors.html -> sponsors/original-sponsors
learning/crystal_programming.html -> learning/crystal-programming
```
`404.html` is explicitly excluded.

Still missing redirects from the old `.html` paths to the new ones to keep existing links working.

Ref: #346


matiasgarciaisaia:
Can you add the blogposts + provide the full list of redirections we should add? And we then maybe sync to merge + deploy the redirections at the same time.

matiasgarciaisaia:
We need to reset the redirections on every deploy since we're using `aws s3 sync` with the `--delete` option (which will delete the redirections every time).

We may be able to hack something to avoid that, but I guess that's OK.

The hard-coded list of URLs (_path-ish's_, actually) is so we only support legacy URLs (instead of allowing future posts to have a .html entry too).

straight-shoota:
So should we merge this? I think it should be good now.

matiasgarciaisaia:
Yes! If you think it looks good, let's merge it :shipit:

PR #359, opened 2022-10-27
Importing the stories into its own space

beta-ziliani:
- [x] Nikola
- [x] Diploid
- [ ] Sikoba


straight-shoota:
I'm not sure they should be a separate content type. I'd rather put them into the existing `post` type and use [post categories](http://jekyllrb.com/docs/posts/#categories) for distinction between success stories and other types of posts. IIRC we talked about this and you confirmed that.

beta-ziliani:
So IIUC I should create them under `stories/_posts`, right? Wouldn't it make sense to do the same with releases?

beta-ziliani:
alternatively we could use tags instead and have them all in the same `_post` folder (same with releases)

straight-shoota:
No, put them in `_posts/` (or `_posts/stories`) so they become part of the posts content type. I think they should appear as articles on the blog, that's easier when they are the same content type.
If you put them in ``_posts/stories` or set `category: stories` in the frontmatter, they belong to the `stories` category. We can then use that category to filter for stories where we want only stories.

I'd suggest to maybe use `success_stories` as category name instead, it's more explicit.

beta-ziliani:
I'm realizing that a problem with merging it into the posts is that you can't navigate them when coming from the success stories section

straight-shoota:
It's possible. But it's also a question of what we want. Prev/next navigation within the category or overall.
Maybe a better approach for navigating between success stories is actually not prev/next but to have a list of recommended next reads (we can fill that manually or automatically, based on similarity for example).

straight-shoota (approved):

PR #362, opened 2022-11-03
Release 1.6.2

beta-ziliani:


straight-shoota (approved):

_releases/2022-11-03-1.6.2-released.md
```diff
@@ -11,6 +11,8 @@ We are releasing the second patch release of the [1.6](/2022/10/06/1.6.0-release
 Pre-built packages are available on [GitHub Releases](https://github.com/crystal-lang/crystal/releases/tag/1.6.2) and our official distribution channels.
 See [crystal-lang.org/install](https://crystal-lang.org/install/) for installation instructions.
 
+**News for `brew` users!** From now on, the interpreter is built by default 🎉
```
straight-shoota, line 14: Not sure if this should be mentioned in the release notes. It's not really related to the release (if it was. I'd rather make a separate post about that, actually.

beta-ziliani: It's a change in the release process, I don't see why it shouldn't be mentioned here. We can make a blog post about it, but it would only be that line... 🤷 I rather update the `install` page

straight-shoota: It's a change in a third party package manager 😛 We just happened to submit a patch to it in association with updating the formula for the 1.6.2 release (and it's not even accepted yet, btw.).

I imagine the blog post to also expand on why the interpreter is *not* enabled in our tarballs and system packages. Without further explanation it seems quite random that you get a build with interpreter from brew, but not in any of our other packages (not even the darwin tarball at GitHub Releases).

PR #369, opened 2022-11-24
Refreshing code snippets

beta-ziliani:
This PR adds and changes the code snippets shown on the web page in order to make them executable in carc.in. Additionally, it fixes the error output that was not showing.

Missing bits:
 * It would be nice to add more examples, in particular about functional programming and calculations. But maybe that's too much?
 * Add  ◀ ▶ controls to keep the list of snippets in one line.
 * The yaml formatting in Dependencies is breaking the indentation for some reason 🤷


straight-shoota:
It's super confusing an a complete mess that the example code is separated from the associated description and metadata. Reviewing that is a pain.

We should merge them together. I think a good solution would be to create a new collection for samples and iterate them in the homepage header.
I can take care of that if you agree.

It's important to keep the number of prominent examples low. Not more than seven. Ideally five. More is not always better. It just becomes convoluted. Let's pick the best ones and polish them up to shine the best. We can have more examples about other use cases in some other location if we want. But the prime examples on the home page header should be concise and clear.

beta-ziliani:
> a good solution would be to create a new collection for samples and iterate them in the homepage header.
> I can take care of that if you agree.

I agree! 💯 

> It's important to keep the number of prominent examples low. Not more than seven. Ideally five. 

I came here to link it from an issue description to create the Perks page, that might have more examples. Then, we can simply have a field to select which ones go into the home page. WDYT?

straight-shoota:
I would rather defer the Perks page. I don't think it's critical. Let's focus on getting the relaunch ready with the current scope. There's still a bunch of work left for that. Once we have that launched, we can come back to add more pages.

beta-ziliani:
There are two issues now:
 1. Type system's description doesn't show up.
 2. Cards are ordered alphabetically.
 
 I don't know what's causing 1. For 2 we can rename the sample files with NN-name.md

straight-shoota:
Resolved by #591

PR #370, opened 2022-11-24
Extending FAQ info in sponsors

beta-ziliani:
Bringing back some missing information. TODO: Ideally, the FAQ should be collapsed instead of shown by default.


beta-ziliani:
This is old news now

PR #373, opened 2022-12-02
Post: Functional Crystal

beta-ziliani:


_posts/2022-11-11-functional-crystal.md
```diff
@@ -0,0 +1,428 @@
...
+    in IntLiteral
+      self.number.to_s io
+    end
+  end
+end
+```
+
+Note how this method is more concise than the OOP one (there's a little white lie, see the coming section). Also importantly, note how we're casing on the type of `self`.
+
+## The missing bits
```
straight-shoota, line 347: I think for this section it would be important not to state the status quo, but also express at least some ideas for future developments. Even stating that some aspect may be unlikely to change would at least give an expectation.

This can definitely be strongly opinionated! This might put you in a bit of a tough spot because as team lead you might be inclined to make realistic predictions😅 . But it is important for the article to give some outlook on these matters.

_posts/2022-11-11-functional-crystal.md
```diff
@@ -0,0 +1,428 @@
...
+      self.operator.to_s io
+      self.right.inspect io
+    in IntLiteral
+      self.number.to_s io
+    end
+  end
+end
+```
+
+Note how this method is more concise than the OOP one (there's a little white lie, see the coming section). Also importantly, note how we're casing on the type of `self`.
```
straight-shoota, line 345: Hm, this part makes me feel a bit uneasy. Sure internalizing the reference to `io` removes side effects that the visitor pattern uses. But then you're just trading multi dispatch for exhaustive case, which might be fine for this simplified use case, but when scaling that up it will lead to complex code in a huge `case` expression.

I don't have a better example at hand, but maybe we can find something more suitable to present pattern matching? (maybe also consider `case` with a tuple for that?)

_posts/2022-11-11-functional-crystal.md
```diff
@@ -0,0 +1,428 @@
...
+an instance variable whose type is, eventually, this same
+struct. This makes it impossible to represent the struct
+in memory, because the size of this instance variable depends
+on the size of this struct, which depends on the size of
+this instance variable, causing an infinite cycle.
+```
+
+A missing piece is that structs are placed in the stack, unlike classes, and that is why the compiler needs to know ahead of time the exact size of it.
+
+If we want to have immutable classes, that are placed in the heap instead of in the stack (and therefore, can have recursive instances), the compiler should add support for it. One way to achieve that is to have the compiler call a method when passing `Reference` objects around, and have `Reference` return `self` and a new `ImmutableReference` type of object return a copy. If it sounds costly, fear not as the LLVM backend should optimize it in `--release` mode.
```
straight-shoota, line 405:
```suggestion
If we want to have immutable classes, that are placed in the heap instead of in the stack (and therefore, can represent recursive structures), the compiler should add support for it. One way to achieve that is to have the compiler call a method when passing `Reference` objects around, and have `Reference` return `self` and a new `ImmutableReference` type of object return a copy. If it sounds costly, fear not as the LLVM backend should optimize it in `--release` mode.
```

I don't get this idea about a compiler feature. If you want an immutable class type, you can get that by just not defining any setter or mutating its instance variables in any other way.
This requires discipline when defining types, which might be sub-par. But it's definitely possible to get immutable types even without any additional compiler features.

beta-ziliani: Not sure I follow. Of course you can build stuff for each particular case, but the point of a compiler should be to make it easier for you to code certain patterns. This said, I now disagree with my proposed solution and more so to the fact that I even added a solution. That's not the point of this article 🤦

_posts/2022-11-11-functional-crystal.md
```diff
@@ -0,0 +1,428 @@
...
+In Crystal, primitive types like numeric types and `struct`s are passed by value.
+
+<!-- If we want to be able to alter the value of `y`, we need to pass it by reference. In Crystal this is doable, but to do it right you need to _box_ it. We won't dicuss this -->
+
+Consider now the following code:
+
+```cr
+def g(arr : Array(Int32))
+  arr[0] += 1
+  arr = [2]
```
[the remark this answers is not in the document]
beta-ziliani: > I'd suggest to maybe leave out the thing about the return value.

I don't follow. There's no discussion about the returned value.

_posts/2022-11-11-functional-crystal.md
```diff
@@ -0,0 +1,431 @@
+---
+title: Functional Crystal
+summary: A tour on the functional aspects of Crystal
+author: beta-ziliani
+---
+
+If someone posts a claim "_Crystal is a functional language_", they would be given a grim look. I mean, classes and inheritance are all over the place in the stdlib. Surely it's an object oriented language, right?
```
a-alhusaini, line 7: This can be shortened to: "if someone claims ..."

_posts/2022-11-11-functional-crystal.md
```diff
@@ -0,0 +1,431 @@
...
+
+If we observe modern programming languages, we'll find that neither functional languages are strictly functional, nor object oriented ones are strictly object oriented. Or, to put it differently, these two concepts are not in contradiction; they can both be part of the tools provided by languages to abstract code. And Crystal is not an exception!
```
a-alhusaini, line 9: "observe" can be simplified to "look at".

"Or, to put it differently" is an unneded sentence here.

_posts/2022-11-11-functional-crystal.md
```diff
@@ -0,0 +1,431 @@
...
+
+But why should you care? As it turns out, functional programming got fancy these years for several reasons, which can be simplified as:
```
a-alhusaini, line 11: "Which can be simplified as" should be removed.

_posts/2022-11-11-functional-crystal.md
```diff
@@ -0,0 +1,431 @@
...
+
+**<center>Functional programming helps write safe and modular code.</center>**
+
+In this post, we will revisit what makes functional programming exciting, and how Crystal helps you write functional code. In particular, we'll talk about three key components of functional programming:
+
+ 1. Immutability.
+ 2. Algebraic data types.
+ 3. Closures.
+
+While discussing each of these, we will point out at some improvements that could be made to the compiler and stdlib to improve its support. The intention is not to fix an agenda, but rather to open the discussion about wether such improvements are relevant to the community.
```
a-alhusaini, line 21: "we will point out at some"

"at" is a typo here.

_posts/2022-11-11-functional-crystal.md
```diff
@@ -0,0 +1,431 @@
...
+
+## Mutability considered harmful
+
+A key aspects of functional programming is related to the idea that any change in the state of the program should be _local_. A way to describe this is with the following comparison, where `f` is any function or method, and `x` any value:
```
a-alhusaini, line 25: "aspects" shold be single.

_posts/2022-11-11-functional-crystal.md
```diff
@@ -0,0 +1,431 @@
...
+end
+
+def f(x)
+  Counter.count
+end
+```
+
+Changing (_mutating_) the count at each call of `f` is a _non-local effect_, as it doesn't depend on exclusively on `f`'s arguments. You could say that the example above is tricky, but I'm certain that you've experience before the surprised of calling a method and getting an unexpected result.
+
+Several languages that consider themselves functional allow such _non-local effects_. These languages are sometimes mentioned as _multi-paradigm_ or _hybrid_, and an important part of this post is to show that Crystal is also such an hybrid language. It definitively favors OOP, but it also allows important functional patterns.
```
a-alhusaini, line 53: Several can be removed. Start with Languages

a-alhusaini:
I really like how this post is coming together! I suggested a few changes just now and will come back and add more in a few hours

_posts/2022-11-11-functional-crystal.md
```diff
@@ -0,0 +1,431 @@
...
+
+### Little diversion: the `record` macro
+
+An alternative implementation of the `Point` struct is using the macro [`record`](https://crystal-lang.org/api/1.6.2/toplevel.html#record%28name%2C%2Aproperties%29-macro). This macro further enhances the experience of working with immutable structs.
+
+```cr
+record Point, x : Int32 = 0, y : Int32 = 0
+```
+
+`record` generates an identical struct as above, together with a handy method `#copy_with` that allows to return a copy of the struct with some given instance variables modified. For instance, we can solve Exercise 2 above defining the `translate_x` method directly when defining `Point`:
```
HertzDevil, line 141: If the above example uses `property` instead of `getter` then theseare not identical structs, as only the one above is mutable

beta-ziliani:
I stumble too many road blocks. Closing this for now.

PR #409, opened 2023-02-10
Update font definitions

straight-shoota:
* Use smaller font files
* Add italic styles
* Remove unused weights


beta-ziliani:
The monospace version has no replacement, so I think it's falling to a default font.

This PR:
image

Original:
image

straight-shoota:
I missed the mono variant. Fixed.

PR #414, opened 2023-03-02
Reset design basics

straight-shoota:
This is the first step towards a website relaunch. It cleans out a lot of cruft in the template and starts with a very basic blank slate. All stylesheets are thrown out and minimal skeleton is implemented.
In this state, the site is not really usable, but it creates the foundation for following updates the focus on individual components.


beta-ziliani (approved):
It's hard to tell where this is going, but I trust you

PR #417, opened 2023-03-06
Update style guide for relaunch

straight-shoota:
Updates the styleguide components for the relaunch.

Note that *Post Teaser* and *Team Member* are left out, they'll be updated on their own later.


beta-ziliani (approved):
This is a much improved page to what we have now 👏

PR #418, opened 2023-03-07
Add devenv with pre-commit configuration

straight-shoota:
DX improvement.
Especially the pre-commit hooks are very helpful to ensure proper formatting. Currently it's only enabled on select file types because prettier has some issues, for example with liquid templates in HTML/Markdown files.


beta-ziliani:
Does this add something to the dev environment worth mentioning in the README?

straight-shoota:
Ah yeah, good point. I didn't even realize there's a `nix-shell` as well. It works a bit different from `devenv` which is pretty new but has lots of nice features.

beta-ziliani (approved):
(Note: I haven't tried locally)

PR #430, opened 2023-03-15
Adjust base font-size to scale based on viewport height as well

straight-shoota:


beta-ziliani (approved):
Looks good. One thing I noticed is that the bar is not stretching right, probably because the icon and name are not stretching.

image

PR #451, opened 2023-06-07
Change public key location in install.sh

beta-ziliani:
Obs changed the location of the key:
https://github.com/crystal-lang/crystal/issues/13552


straight-shoota:
This change needs to be applied to https://github.com/crystal-lang/distribution-scripts/blob/master/packages/scripts/install.sh as well.

PR #459, opened 2023-06-15
[CI] Fix ownership and parameterize output directory

straight-shoota:


.github/workflows/deploy.yml
```diff
@@ -40,10 +41,11 @@ jobs:
       - name: Build website
         run: |
           docker compose run --rm web /bin/bash -c "bundle check || bundle install --jobs=3"
-          docker compose run --rm web make build
+          docker compose run -e O --rm web make build
```
beta-ziliani, line 44:
```suggestion
          docker compose run -e $O --rm web make build
```

beta-ziliani (changes requested):
Why "O"?

.github/workflows/deploy.yml
straight-shoota: No, that's correct. `-e O` means to pass the environment variable `O` into the container. It's short for `-e O=$O`.

straight-shoota:
`O` is the name for the target directory in `Makefile`. It seems easier to just utilize that name as well outside of it. It's established for that.

beta-ziliani (approved):

PR #472, opened 2023-07-18
Update install instructions for Windows

HertzDevil:


straight-shoota:
Yes, I think `scoop` is certainly useful. And `winget`, sure.
They're pretty much comparable to homebrew and we wouldn't want to remove that as well.

straight-shoota (approved):

HertzDevil:
The Homebrew packages are officially maintained, the Scoop and Winget ones aren't (and the package definitions cannot update themselves). So IMO we risk suggesting people to install outdated versions of Crystal

straight-shoota:
We send patches to the homebrew package, but we don't maintain it.
Sources that are not owned by us are distribution package repositories like BSD ports and packages, apk (Alpine), portage (Gentoo), pacman (Arch Linux).
And most repos for 3rd party package managers: asdf, homebrew, scoop, termux, nix.

PR #476, opened 2023-07-19
Add `/newsletter` redirects

straight-shoota:
This creates a couple of URLs for the newsletter which then redirect to mailchimp. This makes it a bit easier to reference the newsletter links and have the final target configured in a single location.


beta-ziliani:
It isn't working the `/newsletter` redirect, but I can't find why 🤔

beta-ziliani (approved):

PR #481, opened 2023-07-21
Add page: Sitemap

straight-shoota:


beta-ziliani:
I don't understand the split between `Pages` and `events`. Also, wouldn't it make sense to have it more in a tree-like fashion?

straight-shoota:
Pages and events are different content types. Just like posts and releases. But I'm not including those in the sitemap for simplicity because they're already accessible through the blog and releases index pages (which are linked from the sitemap). Once we have an events index page, we can probably remove them as individual entries from the sitemap as well, leaving only pages.

Yeah I guess, a tree would be nice. But there's not really much tree information available. You could only go by URL and that's not very helpful because the path scheme does not represent the information architecture.

PR #486, opened 2023-07-22
Add page: Used in Production (MVP)

straight-shoota:
This is just a minimal implementation to make this page presentable. The entire structure is supposed to be overhauled, but this can happen at a later point.

There's a small glitch in the callout, which is fixed in #485.


beta-ziliani:
Isn't there enough material in the new site to do the new layout instead? (I agree this can be added later though)

straight-shoota:
It's some additional work that I'd like to defer. We can pick this up again before releasing the new site, but it shouldn't hold things up. This is good to go, it makes the page presentable.

PR #480, opened 2023-07-20
Add page: Community

straight-shoota:

PR #487, opened 2023-07-22
Add page: Crystal 1.0 Conference

straight-shoota:
Speaker icons are missing. They are pulled directly from airtable which probably isn't the best idea. This is broken in the current website already and I'm not sure it's worth fixing. It should be a separate change though.

Goes on top of https://github.com/crystal-lang/crystal-website/pull/480


beta-ziliani (commented):
How hard is to add the speakers icons?

beta-ziliani (approved):

straight-shoota:
> How hard is to add the speakers icons?

Dunno. It's a bit of legwork, but likely not complicated. Have to track the images in Airtable. Probably put them in the repo to be future proof. And then replace the URLs in https://github.com/crystal-lang/crystal-website/blob/master/_data/conference_talks.csv

beta-ziliani:
I added it as a polish task for the future. This is good to go 🚀

straight-shoota:
I don't think fixing the images belongs in the scope of the relaunch. It's already broken in the current page: https://crystal-lang.org/events/2021-crystal-1.0/

PR #502, opened 2023-10-06
Add conference sponsors & schedule

straight-shoota:


beta-ziliani (approved):
Icon alignments and the sizes of the button in the text aren't perfect, but we can fix those later.

straight-shoota:
What icons? And what's wrong with button alignment?

beta-ziliani:
sorry, I meant the logos, they aren't vertically centered, are they? And the last button (map) has a smaller font than the rest. Again, nitpicks.

straight-shoota:
The logos are vertically center aligned. Maybe one of the logos if off-centered in itself? I don't think they look unbalanced, though.

The buttons are kinda intended to be different sizes. But we can align them.

PR #509, opened 2023-10-19
Add page: Book: Crystal Programming

HertzDevil:


_pages/learning/crystal_programming.html
```diff
@@ -1,81 +1,124 @@
...
 
-headline: Building efficient, safe, and readable web and CLI applications with Crystal
-cover_image: /assets/learning/crystal_programming.png
-vendors:
-- url: https://www.amazon.com/Crystal-Programming-project-based-introduction-applications/dp/1801818673/ref=sr_1_2
-  name: Amazon
-- url: https://www.packtpub.com/product/crystal-programming/9781801818674
-  name: Packt
+<div class="book-presenter">
+  <div class="book-presenter__main">
```
straight-shoota, line 7: The nested main div seems unnecessary.
Other places use a grid layout with template areas for this. E.g. `grid-template-areas: "cover description" "body body"`. This keeps the html minimal and semantic.

_pages/learning/crystal_programming.html
```diff
@@ -1,81 +1,124 @@
...
-  name: Crystal Programming
-  isbn: 978-1801818674
-  publisher: Packt Publishing
-  numberOfPages: 356
-  datePublished: 2022-05-27
+    <div class="book-presenter__description">
+      <summary>
+        <strong>Crystal Programming</strong> is a project-based book by George Dietrich and Guilherme Bernal that covers all the
+        fundamental concepts needed to start programming in Crystal, such as OOP and concurrent programming.
+      </summary>
```
straight-shoota, line 16: This is not correct usage of the `<summary>` element. It's only valid inside a `<details>` element. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary

If your intention is to make semantic annotations: I don't think we have much of that currently, but I'd appreciate it very.
This can be expressed using microdata for https://schema.org/Book (this needs to go along with an itemscope definition and other metadata of course).
```suggestion
      <p itemprop="abstract">
        <strong>Crystal Programming</strong> is a project-based book by George Dietrich and Guilherme Bernal that covers all the
        fundamental concepts needed to start programming in Crystal, such as OOP and concurrent programming.
      </p>
```

straight-shoota (commented):
Looks good overall.

What's the reason for removing the layout, though? I suppose it might be ill-named as `learning`, but it's functionally very relevant. We expect there to be more pages about books or possibly other resources and they should all share the same layout. Of course we can also copy and paste, but Jekyll layouts are great for this kind of application.

I don't think there should any page-specific styles, though.  Everything it uses can be used in other contexts at well. The style classes would fit very well as general elements or component, respectively. (I think `hex-icon` should be considered an element, the rest components).
Corresponding examples for these style features should be added to the style guide.

A nitpicking comment on `image-list`: Semantically it's not ideal to place icons and text on the same level in the list container. An ideal solution would place pairs of icon and text into an item container, which is then nested in the list container.
I'm happy to accept it as is though. The argument is more philosophical and has little practical impact. The current solution is entirely functional sufficient.

_sass/components/_image-list.scss
```diff
@@ -0,0 +1,25 @@
+.image-list {
+  display: grid;
+  grid-template-columns: repeat(1, min-content 1fr);
+  align-items: center;
+  row-gap: 20px;
+
+  .small {
+    font-size: 85%;
+    margin-right: var(--padding-lg);
```
straight-shoota, line 9: **nitpick:** This results in an odd margin on the rightmost column:
grafik

I think the only way to solve this properly is to merge the icon and text elements into a single grid entry. Then you can use `column-gap` to have a spacing only between columns, but not at the outer edge.

_sass/components/_image-list.scss
```diff
@@ -0,0 +1,25 @@
...
+    grid-template-columns: repeat(2, min-content 1fr);
+  }
+}
+
+@media (min-width: 55em) {
+  .image-list {
+    row-gap: 60px;
+    grid-template-columns: repeat(3, min-content 1fr);
+  }
+}
```
straight-shoota, line 25: **suggestion:** I think these explicit media queries should be unnecessary. The entire layout can be configured to be implicitly adaptive.

* `row-gap` can be an adaptive value such as `var(--block-flow-sm)`.
* `grid-template-columns` can be adjusted implicitly with something like `repeat(auto-fill, minmax(30ch, 1fr))`.

_sass/elements/_hex-icon.scss
```diff
@@ -0,0 +1,43 @@
...
+      100% 75%,
+      100% 25%,
+      50% 0,
+      0% 25%
+    );
+
+    & > * {
+      position: absolute;
+      width: 24px;
+      height: 24px;
```
straight-shoota, line 37: **nitpick:** Looks like these values could be expressed as functions of `--icon-height`.
```suggestion
      width: calc(var(--icon-height) / 2);
      height: calc(var(--icon-height) / 2);
```

_sass/elements/_hex-icon.scss
```diff
@@ -0,0 +1,43 @@
...
+  --sin-60: 0.866;
+}
+
+.hex-icon {
+  --icon-height: 48px;
+
+  width: calc(var(--icon-height) * var(--sin-60));
+  height: var(--icon-height);
+  display: inline-block;
+  margin-right: 12px;
```
straight-shoota, line 11: **nitpick:** This margin is odd. It doesn't seem to be part of the icon itself but depend on the context it's placed in.
Removing that aligns with the above suggestion to combine icon and description into an intermediary container. This whitespace should be applied to that inner container.

_layouts/book.html
```diff
@@ -0,0 +1,24 @@
...
+<section>
+  <div class="full-width-rule">
+    <div class="hex">
+      <span>What you will learn</span>
+    </div>
+  </div>
+  {% include components/image-list.html items=page.what_you_will_learn %}
+  <div class="full-width-rule">
+  </div>
+</section>
```
straight-shoota, line 16: **issue:** I think this section needs more horizontal spacing between its elements (compare invision design). `--block-flow: var(--block-flow-md)` would do, but that also affects the sections own top margin.
We could introduce a custom class for a section with bigger block spacing.
However, I think we're just missing a general rule for `.full-width-rule`. The blockflow mixing already has some additional spacing around `hr` and the same should probably apply to `.full-width-rule`.
So the diff for that would be:
```diff
--- i/_sass/mixins/_blockflow.scss
+++ w/_sass/mixins/_blockflow.scss
@@ -5,13 +5,13 @@
     + * {
       margin-block-start: var(--block-flow, $width);

-      &:is(hr) {
+      &:is(hr, .full-width-rule) {
         margin-block-start: var(--block-flow-md);
       }
     }
   }

-  :where(& > hr + *) {
+  :where(& > :is(hr, .full-width-rule) + *) {
     margin-block-start: var(--block-flow-md);
   }
```

_sass/components/_book-authors.scss
```diff
@@ -0,0 +1,31 @@
...
+  grid-template-columns: auto 1fr;
+  column-gap: var(--padding-sm);
+
+  > * {
+    grid-area: main;
+  }
+
+  > p {
+    max-width: 43ch;
+  }
```
straight-shoota, line 23: **question:** Is there a particular reason for this restriction? It seems quite arbitrary. I think the layout would be better without it.

_sass/components/_book-authors.scss
```diff
@@ -0,0 +1,31 @@
...
+
+  > .book-author-icon {
+    grid-area: icon;
+    width: 6ch;
+    height: 6ch;
```
straight-shoota, line 28: **suggestion:** `aspect-ratio: 1` is more explicit about the intention. And adding `object-fit: contain;` would future-proof for images that are not square.
```suggestion
    width: 6ch;
    aspect-ratio: 1;
    object-fit: contain;
```

We should probably have an abstraction for a rounded image element. There are several ones already in the codebase. Feel free to introduce that, or I'll take care of it afterwards.

straight-shoota (approved):

PR #512, opened 2023-11-02
Removing header for conference

beta-ziliani:


straight-shoota:
Haha, I was thinking about that as well. I'd rather leave the file in place though and just disable the include. That makes it easier to pick up when we need it again.

PR #527, opened 2023-11-17
Use manas SVG logo

straight-shoota:
It's the same image, just a different format.


beta-ziliani:
This is machine-generated from https://github.com/crystal-lang/crystal-website/blob/master/_data/others.json, so it's there where the change should happen.

I will close this one.

beta-ziliani (approved):

PR #532, opened 2023-11-20
Add YouTube channel to community page

straight-shoota:


beta-ziliani:
youtube.svg file missing

beta-ziliani:
Doesn't it make sense to add https://www.youtube.com/@manastechnologysolutions8418, given it contains many videos from Crystal's sphere?

beta-ziliani (approved):

straight-shoota:
@beta-ziliani I'd rather make those videos accessible from the Crystal Channel. In fact, the recordings from Crystal 1.0 are already available there. Seems like I missed the live streams because I was only looking at the videos.

PR #534, opened 2023-11-23
Add footer

straight-shoota:
This is a propsal for a footer with a bunch of navigation menus. This allows for great cross-navigation from anywhere on the site to the most important destinations, without having to go through the full site hierarchy.

The footer also includes <del>an edit link to facilitate contributions from readers (#92)</del><ins>(moved to separate PR)</ins> and a newsletter signup form.

grafik


beta-ziliani:
I love this! there's a little detail, but let's hear it first from the design team.

straight-shoota:
This PR was previously depending on #529 but the overlap is minimal, so I rebased it onto the relaunch branch directly.

It may need some visual adjustments, but I think it's pretty much ready for review.

_includes/site/footer.html
```diff
@@ -1,17 +1,133 @@
...
-    <div class="col s12 m12 l6 manas right-align black-text">
-      Crystal language, born & raised at
-      <a href="https://manas.tech" target="_blank" class="black-text">Manas</a>
-      <a href="https://manas.tech" target="_blank" class="logo">
-        <i class="manas"></i>
-      </a>
-    </div>
-  </div>
-</footer>
+<nav class="footer-section" aria-label="Home section links">
```
beta-ziliani, line 1: This isn't working well: it's taking the `home` list for all the lists.

Also, would it be over-engineering it to have an outer iteration for the lists, given that everything in each `<nav>` element is the essentially the same?

_includes/site/footer.html
```diff
@@ -1,17 +1,133 @@
...
+<nav class="footer-section" aria-label="Community section links">
+  <a class="link-item__main" href="/community">Community</a>
+
+  <ul role="list">
+    {% assign nav = site.data.nav.community %}
+    {% include site/nav_links.html links=nav %}
+  </ul>
+</nav>
+
+<div class="link-items">
```
beta-ziliani, line 26: Ditto, would it make sense to have this configurable as well?

_includes/site/footer.html
```diff
@@ -1,17 +1,133 @@
...
+
+<div class="newsletter-footer">
+  <form
+    action="https://tech.us6.list-manage.com/subscribe/post?u=68338ca88d5d6326c9a824317&amp;id=6f80e8b0be&amp;f_id=005a33e3f0"
+    method="post" id="mc-embedded-subscribe-form" name="mc-embedded-subscribe-form" target="_self"
+    novalidate="">
+    <p>Subscribe to our newsletter</p>
+      <input
+          type="email" name="EMAIL" class="required email" id="mce-EMAIL" required="" value="" placeholder="E-Mail">
+      <button type="submit" name="subscribe" id="mc-embedded-subscribe" class="btn btn-flat btn-large hex">Submit</button>
```
beta-ziliani, line 96: this is a huge button. I either make it vertically centered w.r.t. the hole form, or have it occupy the same vertical space as the input box.

_includes/site/footer.html
```diff
@@ -1,17 +1,133 @@
...
+      <div style="position: absolute; left: -5000px;" aria-hidden="true">
+        <input type="text" name="b_68338ca88d5d6326c9a824317_6f80e8b0be" tabindex="-1" value="">
+      </div>
+  </form>
+</div>
+
+<div class="link-items link-items--social">
```
beta-ziliani, line 103: Again, shouldn't this be configurable?

beta-ziliani: Also, they location is a bit off but I can't provide an alternative layout as a proposal.

beta-ziliani (changes requested):
There's a bug that needs fixing, the rest are questions but not requirements.

_includes/site/footer.html
straight-shoota: I don't follow what you mean. Could you elaborate?

Yes, I think that would be over-engineered. It makes it harder to move stuff around. A little duplication isn't bad.

beta-ziliani: > It makes it harder to move stuff around.

I don't get how, but OK, it was a question and not a request after all.

straight-shoota:
There was a little hiccup with the `nav_links` include which lead to the same set of links (a completely wrong one!) being used in all lists.
Probably caused by pulling this out from a longer branch.

Fixed now:

grafik

_includes/site/footer.html
straight-shoota: The form will need some visual enhancements. I want to merge the big PRs first to get everything together. And then we can polish the details.

_includes/icons/edit.svg
beta-ziliani, line 1: Does this belong to this PR?

beta-ziliani (approved):

_includes/icons/edit.svg
straight-shoota: No. That's for #559

PR #536, opened 2023-11-24
Standardize on `page_title` property

straight-shoota:
In the same vain as #524 this unifies the `header_title` and `page_title` properties into the latter (which is used more commonly).


beta-ziliani:
Shouldn't it be changed in `/_includes/site/header.html`? Grep says there's a `header_title` there still

straight-shoota:
I don't think that include is used anywhere anymore. It should be removed entirely.

beta-ziliani:
then, like, do it? :-) I mean, why not here where we find out?

straight-shoota:
Because it's unrelated to this particular change 😛 
I'm already collecting for another cleanup round of unused files.

beta-ziliani (approved):

PR #543, opened 2023-12-13
Add `jekyll-relative-links` plugin

straight-shoota:
This plugin enables markdown links to refer the *source path*, instead of the URL of the rendered page. These paths are implicitly converted to the permalink paths, so the final result in the generated site is identical.

The result is that markdown links work on the source files directly, independently of Jekyll.

For example, when you view the preview of [`_pages/team.md` on GitHub](https://github.com/straight-shoota/crystal-website/blob/3d448c688b7a335800accf09f355f821f7f3f928/_pages/team.md), or [visit the generated *Team* page](https://deploy-preview-543--crystal-website.netlify.app/team/), clicking on the *Governance document* link, it takes you to the right place in both cases.

This is a tiny quality-of-life improvement, not critical but makes some things easier.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ gem 'kramdown-parser-gfm'
 group :jekyll_plugins do
   gem 'jekyll-seo-tag'
   gem 'jekyll-redirect-from'
+  gem 'jekyll-relative-links'
 end
 
 # Webrick is used for `jekyll serve`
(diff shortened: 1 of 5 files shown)


beta-ziliani (commented):
Cool. Surely other links should be updated too to take this change, right? The two presented here are just examples, I imagine.

beta-ziliani (approved):

straight-shoota:
> Cool. Surely other links should be updated too to take this change, right? The two presented here are just examples, I imagine.

A limitation is that it works only with markdown links. And most of the markdown content doesn't link internally. Instead, most of the internal links (e.g. on the community page) are written in HTML.
We're definitely progressing more towards Markdown (which is a good thing on its own), and I expect more Markdown content in the future with internal links.
Also there might be more internal links that I missed, those two were just very obvious ones that I came across..

PR #546, opened 2023-12-15
Add `copy-action` component

straight-shoota:

PR #547, opened 2023-12-18
Feat/install mobile

jkicillof:
Ii took @straight-shoota `feat/install` as starting point:
* Added style to the Instructions button and included the repo badge inside
* Replaced the svg icons with Google Material Symbols
* Fixed repo badge colors
* Removed extra white space
* Added copy button

**Desktop**
image

**Mobile**
image


straight-shoota:
Notes from off-site communication:

* Poor task synchronization has lead to this and #540 evolving in parallel. We'll have to merge the different changes together.
* The button appliance for instructions doesn't stand out very much. It has a thin border, which doesn't appear too strong next to the row borders. The chevron icon is detached from the text label. I'd suggest to go with a normal link style (underlined) as in #540. It *could* enhance the button style, but IMO the button isn't really necessary and creates too much visual difference between titles that are links and those that are not.
* Nesting links inside each other (title/instructions and repology badge) isn't a great solution. It only works with a button + JS hack as HTML already prevents you from nesting anchors. I think we should not try to fight this and instead embrace non-nesting as a design contraint.
* So far we've been exclusively using SVG icons in the relaunch codebase. I don't think there's much gain from converting them to font icons. At least not as a singular byproduct of one page. If we want to reconsider the decision SVG vs icon fonts, we should have a dedicated discussion about that and standardize on one technology site-wide.
* The copy button is duplicate to #546.

straight-shoota:
I incorporated the relevant visual changes from this PR into #540 and #546.

PR #548, opened 2023-12-19
Redesign header with breadcrumbs

straight-shoota:
This is a design study which merges breadcrumbs from #529 into the site header.
It also takes some inspiration from the current website placing the Crystal logo in the top left corner of the viewport.

The result is a very lightweight header with efficient in-section navigation. It scales well and works great on all screen sizes (scaling is not yet completely polished, though).

It's not active on the home page, so you need to go to any other page. The home page has a different nav which still needs styling according to the home page designs.

Cross-section navigation falls a bit short as it is only possible through navigating up the tree. This is enhanced by other features such as the footer nav (#534). And we could consider adding some direct nav elements to top bar (example: https://github.com/crystal-lang/crystal-website/pull/529#issuecomment-1825725762).

grafik


jkicillof:
I think this doesn't look as important as a nav bar, my proposal has the same functionality but includes clear hierarchies. We could iterate this if some styles doesn't look right to you.

image

beta-ziliani:
I like the idea to have a back button to return to the main menu in the current page. Otherwise the only way to go visit another page is going to the homepage.

beta-ziliani:
Alternative: can we keep the main menu atop and the breadcrumb? This is what they do in Kotlin ([example 1](https://kotlinlang.org/community/user-groups/)), with the menu being on top as usual, the name of the current section to the left and the menu with subpages on the right. In a sense, this would be similar to #529 but with the menu to the right:

image

`Home` then does not make sense, so maybe the breadcrumb is not that; it's just the title of the section and the menu on the right contains the current page marked (with Overview being the default for the section?). Again, this are ideas from Kotlin, so looking at that page might explain better what I mean ([example 2](https://kotlinlang.org/community/)).

straight-shoota:
Yeah, I had considered this navigation contept. But I don't think it works well with our information architecture. It works best for a relatively small and flat hierarchy where every page is a direct sub-page of one of the items in the top navigation.
Such a structure is very valuable because it's very approachable and easy to use and implement.
Our structure however is more complex with more sections and nesting.

The [sitemap in this branch](https://deploy-preview-548--crystal-website.netlify.app/sitemap/) shows this very clearly. It lists all pages by their section. There are way more sections than we could possibly fit into the top menu. Maybe not every section needs to exists and we could reschuffle a bit, but I doubt that would get it work.
Note that this doesn't represent the entire hierarchy. Some sections are supposed to be sub-sections (e.g. Community > Events).
This expresses very nicely with the combination of breadcrumbs and child pages. It allows navigation upwards and downwards the tree.

beta-ziliani:
My issue is that by having just the breadcrumb you lost connection to the rest of the pages that needs more attention

straight-shoota:
Yeah, but I think it's a problem that breadcrumbs and side nav are visually distant.

> My issue is that by having just the breadcrumb you lost connection to the rest of the pages that needs more attention

I don't follow what you mean here. Can you clarify?

straight-shoota:
I have added the main nav with direct links to the most important pages, oriented to the right of the header near the already existing short-cut links with icons.
This completes the original idea from https://github.com/crystal-lang/crystal-website/pull/529#issuecomment-1825725762, as mentioned in the OP.

beta-ziliani:
YES!

straight-shoota:
So I think this is ready now.

grafik

_posts/2023-07-19-Remilia-interview.md
beta-ziliani, line 1: This shouldn't go here

_style_guide/components/breadcrumbs.html
```diff
@@ -0,0 +1,37 @@
+---
+title: Breadcrumbs
+type: components
+module: _includes/site/breadcrumbs.html
+
+examples:
+  community:
```
beta-ziliani, line 7: The links are being displayed with underline, which is not the right format.

_style_guide/components/breadcrumbs.html
```diff
@@ -0,0 +1,37 @@
...
+    section: community
+    title: Foo Bar
+    url: /foo/bar
```
beta-ziliani, line 10: Is it right that this is outputted exactly as the next one?

_data/nav.yml
```diff
@@ -55,3 +63,31 @@ home:
   url: /releases/
 - label: Sitemap
   url: /sitemap/
+
+install:
```
beta-ziliani, line 67: THese should come from the new install page, right? Or is this different?

straight-shoota: No, that's different. This navigation is between the sub-pages. The install page itself additionally has a TOC.

beta-ziliani (approved):
This is good to go after merging with the main branch (CI ain't happy!)

I wouldn't worry now, but at some point we might want to look into improving this menu:
image

straight-shoota:
CI will be happy after #577 is merged.

PR #556, opened 2024-01-15
Fix gap on `profile-card` between name and handle

straight-shoota:


beta-ziliani (approved):
Before and after (note the handle):
image
image

PR #557, opened 2024-01-15
Tune link style

straight-shoota:
Link styling has historically been quite confusing with different styles in different places. The current designs also show different style: sometimes underlined black text, sometimes non-underlined light blue. The relaunch branch had followed this. But it's time to standardize on a unique link styling.

Links have the current color and an underline serves as indicator.

This removes the shrill light blue link color which leads to calmer visuals and a smoother reading experience.

Before:
grafik

After:
grafik

### DIFF
--- a/_sass/_utilities.scss
+++ b/_sass/_utilities.scss
@@ -15,6 +15,7 @@
     vertical-align: text-bottom;
   }
 
+  @extend .ui-link;
   --link-color: var(--light-text);
   --link-hover-color: var(--lighter-text);
 }
@@ -47,4 +48,10 @@ li {
   &.no {
     list-style-image: url("/assets/icons/close.svg");
   }
+}
+
+// This class marks a link as being part of the UI chrome, which disables some
+// visuals that are usually found on links. They're not underlined, for example.
+.ui-link {
+  --link-underline: none;
 }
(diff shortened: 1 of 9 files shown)


beta-ziliani:
I'm asking a review from Joni to get his opinion. One thing that I don't love is the way it looks on large titles:
image

straight-shoota:
Yeah, I suppose we could consider dropping the link style on headlines.

beta-ziliani:
Actually, I just don't like it when it spawns two lines, but it looks good in one-liners. Just a nitpick I guess.

straight-shoota:
We'll need shorter titles then 😆

straight-shoota:
Running another idea, I think the underline could also have some color contrast to the font color:

grafik

grafik

straight-shoota:
I went ahead with that. The result is more lightweight.

grafik

grafik

beta-ziliani:
I like the new style better, but I still think the blog will look better without the underline

straight-shoota:
I removed the underline on featured posts.

beta-ziliani (commented):
The post back and forth nav items were always broken in this revamp?

image

straight-shoota:
Yeah, that's a another issue on my list. Independent from this PR.

beta-ziliani (approved):

jkicillof:
I agree on dropping blue but the underline should be only for body text, not headlines, nav or buttons

PR #558, opened 2024-01-15
Add indicator for external links

straight-shoota:
We have an indicator for external links already in the top navigation. This expands that to the entire content. The indicator shows wether a link is internal or goes to a different website.
This style can be disabled, for example for links that are part of the UI chrome. Or in contexts where it's clear that a link is outgoing, such as the sponsors list.

Before:
grafik

After:
grafik

This PR goes on top of #557


beta-ziliani:
In the Alumni section it looks a bit scrambled. I think here we might add the GH icon next to the name instead? The rest looks good.
image

straight-shoota:
I'll remove the indicator there.

straight-shoota:
Updated colors after #557

grafik

beta-ziliani (approved):
LGTM. A bit cramped here, but that can be tweaked in a second iteration.

image

PR #561, opened 2024-01-17
Add learning page

straight-shoota:


beta-ziliani:
It'd be better to have a fixed vertical space between items, probably scaling down the book image, and even better add an image to put to the other items (ruby and exercism logos?)

straight-shoota:
I mean: before this PR -> after this PR.

beta-ziliani:
I still think a more direct translation will look better: just remove the book cover from this page. And adding the icons would be a plus to make it look closer to what we want.

beta-ziliani (approved):

PR #562, opened 2024-01-17
Fix post-navigation layout

straight-shoota:
See https://github.com/crystal-lang/crystal-website/pull/557#pullrequestreview-1828028571


_sass/components/_post-teaser.scss
```diff
@@ -38,12 +38,6 @@
   > .content {
     grid-area: content;
   }
-
```
beta-ziliani, line 41: shouldn't also the `footer` grid-area be removed from the template?

PR #565, opened 2024-01-18
Tune footer layout and styles

straight-shoota:
Depends on #564

grafik


beta-ziliani:
similar to the Help section, I think the media channels should not be underlined. I guess a good approximation is to say that the underline goes only in links that are embedded in text.

beta-ziliani (approved):

PR #567, opened 2024-01-19
Add arch/targets to install page

straight-shoota:
This patch adds to the install page for which architecture packages are available.

grafik

### DIFF
--- a/_pages/install.md
+++ b/_pages/install.md
@@ -18,12 +18,18 @@ repository.
     <div class="install-entries">
       <div class="install-entry">
         <span class="title"><a href="on_linux#installer" title="Instructions for Linux installer">Installer (DEB &amp; RPM)</a></span>
+        <span class="targets">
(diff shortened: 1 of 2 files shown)


beta-ziliani:
what's the difference between "universal" and "x86_64 aarch64"?

I'm not convinced of the visuals, like the gap it generates on the entire line. Placing it to the name as before maybe?
image

beta-ziliani:
Although it might make the line super lengthy. It's already longer than the viewport

straight-shoota:
Yeah, no it doesn't work to put everything in one line.
The additional vertical gap isn't that much actually. So not much space is lost. And I actually like it. It's more breathing space to separate the individual items on this very dense layout.

> what's the difference between "universal" and "x86_64 aarch64"?

The former means a single package for both architectures. This is identical to how we label the release artifacts.

beta-ziliani (approved):

PR #570, opened 2024-01-22
Extract package data for install page

straight-shoota:
Extract all the package information from a big HTML blob into structured data and a template to render that.
This is overall more succinct and easier to work with. Data and presentation are separated. It also makes it possible to re-use some of the information in other place (for example in the install sub pages).

Should've probably done this a long time ago.

Alternatively we could use a new custom collection for package data. Would be pretty much the same result except that the content of packages.yml` is spread across individual files for each entry. But I think it's a bit easier to manage in a single data file.

The generated HTML changes slightly because we're now using standard markdown code blocks instead of `{% highlight %}` tags. This required some adjustments in the styles. There are (supposed to be) no visual changes however.

### DIFF
--- a/_data/packages.yaml
+++ b/_data/packages.yaml
@@ -1,0 +1,295 @@
+- type: crystal
+  os: Linux
+  arch: [x86_64]
+  title: Installer (DEB &amp; RPM)
+  label: Linux installer
+  instructions_href: on_linux#installer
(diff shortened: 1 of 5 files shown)


beta-ziliani (approved):
I was so much hoping for this follow up PR <3

PR #573, opened 2024-01-22
Extract `.link-actions` component

straight-shoota:
Refactors `.release-links` into a reusable component.
It just gets a different name and better organization, as well as a style guide entry.
Basically no relevant visual changes, except some a tiny adjustment to make the gap scale with the viewport.


beta-ziliani (approved):
🚀

PR #575, opened 2024-01-24
Add conference pages

straight-shoota:
Updates the conference pages for the new design. It's not super sophisticated, but consistent with the rest of the site. We can iterate on the design later.

grafik


beta-ziliani (approved):
Love the engineering of it 🤩 

We need to remember to add a landing page and the page for Raw Crystal 2020.

PR #578, opened 2024-01-25
Fix nightlies page

straight-shoota:
The nightlies page was broken since #570 because it didn't make the transform.
This PR extracts package data for nightlies and builds the nighlies page in the same was as the install page.
Als improves readability on the normal package data file.

### DIFF
--- a/_pages/install/nightlies.md
+++ b/_pages/install/nightlies.md
@@ -3,134 +3,31 @@ page_title: Nightly Builds
 layout: page-wide
 page_class: page--segmented
 ---
-
 Nightly builds of the Crystal compiler are available from these locations.
 
 ## Linux
 
-<div class="install-panels">
-  <div class="install-group">
-    <h5>Crystal</h5>
-    <div class="install-entries">
-      <div class="install-entry">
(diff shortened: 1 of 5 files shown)


_pages/install/nightlies.md
```diff
@@ -3,134 +3,31 @@ page_title: Nightly Builds
...
-        <a href="https://github.com/neatorobito/scoop-crystal" title="Scoop repository for Crystal on GitHub" class="info">{% include icons/info.svg %}</a>
-      </div>
-    </div>
-  </div>
-</div>
+{% include pages/install/section.html os="Windows" channel="nightly" %}
 
 ## Docker
 
 Nightly builds are available on the `nightly` tag on the [Docker repository of Crystal](https://hub.docker.com/r/crystallang/crystal/).
```
beta-ziliani, line 24: This is now kind of redundant, isn't it?

beta-ziliani (approved):
Shouldn't "nightlies" and "from source" be in the TOC?

_pages/install/nightlies.md
straight-shoota: Yes, it'll be removed in a follow-up.

straight-shoota:
> Shouldn't "nightlies" and "from source" be in the TOC?

Yes, there's a follow-up coming for that.

PR #587, opened 2024-01-26
Add table with all archive download links

straight-shoota:
This is one of the fruits from #570: We can easily display package information in other places.

grafik

### DIFF
--- a/_data/packages.yaml
+++ b/_data/packages.yaml
@@ -189,7 +189,7 @@
   arch: [x86_64]
   title: Installer (`.exe`)
   instructions_href: /install/on_windows/
-  archive_href: https://github.com/crystal-lang/crystal/releases/download/{{ version }}/crystal-{{ version }}-1-windows-{{ arch }}-msvc-unsupported.exe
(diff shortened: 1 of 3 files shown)


beta-ziliani:
Most of the window links are 404. Shouldn't this be caught by CI?

straight-shoota:
Yeah, the link checker should notice that. And I'm sure I had tested the links as well 🤔

Ah so the issue with nightly.link is that it refers to the last run of https://github.com/crystal-lang/crystal/blob/master/.github/workflows/win.yml
This workflow runs on every commit, but it only builds release artifacts on specific commits triggers (on a tag or `ci/*` branch).

I think what we'll need to do is extract the release parts into a different workflow which is only triggered on such occasions. Then the latest build should be compatible with nightly.link.
And we would want this workflow to trigger on a schedule to have actually nightly builds.

beta-ziliani:
Should we put this one on hold until #14256 is solved? Also, do you know why CI didn't catch the error?

PR #593, opened 2024-01-31
Add `@HertzDevil` to `blog_authors`

straight-shoota:
Quinton is missing from the authors data, which renders as anonymous authorshipt ;)

grafik

Of course the template should also fall back to display the handle.


beta-ziliani (approved):
🙈

PR #605, opened 2024-02-01
Add homepage

straight-shoota:
This is the last major missing piece for the relaunch... the home page 🚀 

This PR introduces a new layout with the home page header including the examples section.
The page content itself consists mostly of pre-existing building blocks from includes.

I'm focusing on the major structure, but a couple of details are still missing:

* The examples need another editorial overhaul. This should be easier now that we can see how they fit in the structure.
* Playground integration.
* Colors for code highlighting (related to playground integration).

grafik


_includes/posts/recent_news.html
```diff
@@ -0,0 +1,11 @@
+<div class="featured-posts">
+  {% for post in site.posts limit: 4 %}
+  <div class="post-teaser">
+    <a href="{{ post.url }}" class="post-title">{{ post.title }}</a>
+
+    {% include posts/meta.html post=post %}
+
+    <div class="excerpt">{{ post.summary | default: post.excerpt | markdownify }}</div>
```
beta-ziliani, line 8: I was expecting all this to be just an include, given that it's the same as in the post page. Can be done in a follow up, of course.

beta-ziliani (approved):
Amazing piece of work! 🚀 
(Not that I'm surprised, but still 👏 )

_includes/posts/recent_news.html
straight-shoota: Yeah, that makes a lot of sense.

PR #611, opened 2024-02-02
Remove border around cards

straight-shoota:
Removes the border around card components. This removes visual disturbance and foucses more on the content.

This may be a bit of a controversial change. The card visuals have been - in my understanding - an integral part of the design language for a long time.
But I think it makes a lot of sense to drop the border while keeping all the other aesthetics of the design. This fits very well with the theme of the site that we have reached now with a clean look.

#### Before
grafik

#### After
grafik

As a minor detail, without an enclosure some of the profile link icons might now seem a bit uncoupled from the main part (especially for Jonne and Serdar in the first row). I don't think this is too much of an issue, but we could also tweak the positining and move those icons to a different position and puts the `roles` property with widely varying height on the bottom

The card style is also used for top sponsors and on the media page (the color cards where already borderless). The borderless appearance works there as well:

grafik

grafik

### DIFF
--- a/_includes/media/color_sample.html
+++ b/_includes/media/color_sample.html
@@ -1,4 +1,4 @@
-<div class="card card--borderless">
+<div class="card">
   <div class="color-sample" style="--color-sample-color: {{ include.hex }};{% if include.color == 'Black' %} --color-sample-border: #000;{% endif %}"></div>
   <span class="name">{{ include.color }}</span>
   <code class="low-key">{{ include.hex }}</code>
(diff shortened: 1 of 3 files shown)


beta-ziliani:
LGTM, but I'm deferring it to @jkicillof

jkicillof:
I think we shouldn't use a layout like this without cards, If you want to ditch cards we should go for a table to keep things organized but I think this was ok as it was.

beta-ziliani:
I think both designs are good, but given there's no clear winner I prefer to postpone this discussion for another iteration. Is it OK we close this @straight-shoota ?

PR #614, opened 2024-02-02
Highlight minor releases

straight-shoota:
Adjusts the styling of version numbers on the releases page so that minor releases stand out a bit more than patch releases.

grafik

### DIFF
--- a/_includes/pages/releases/releases-table.html
+++ b/_includes/pages/releases/releases-table.html
@@ -9,7 +9,8 @@
   <tbody>
     {%- assign releases = site.releases | reverse %}
     {%- for release in releases %}
-      <tr {% if forloop.index0 == 0 %}class="latest-release"{% endif %}>
+      {%- assign suffix = release.version | slice: -2, 2 %}
+      <tr class="{% if forloop.index0 == 0 %}latest-release{% endif %}{% if suffix == '.0' %} minor-release{% endif %}">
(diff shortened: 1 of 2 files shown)


beta-ziliani:
but wouldn't it be better to let stand the latest patch of each minor instead? I guess it depends why one is on this page, is to see the release post for a particular minor, or to download the latest patch of a specific minor.

straight-shoota:
Yeah, I suppose that could make sense as well. Since we're typically only patching the latest release, it's naturally always the item below the highlighted minor. 🤷

Anyway, my goal is to highlight the minor releases because they have the most interesting release posts. With many changes and new features. So that's what most readers would be interested in. The patch release notes are typically just something like "we fixed bug X which was introduced in the last minor release".

beta-ziliani (approved):

straight-shoota:
I'm wondering if perhaps we should have a combined release page for every minor branch, including all the changes in any of its patch versions. The use case for this is to easily see all changes between 1.10 and 1.11, for example. With individual release notes you have to look at each 1.11.0, 1.11.1 and 1.11.2 to get the total. Sometimes patch releases revert previous changes, so you cannot be sure that every change from x.y.0 still applies to x.y.z (for example, 1.11.1 reverted the addition of `ReferenceStorage` from 1.11.0).

PR #620, opened 2024-02-05
Improve colors for code highlighting

straight-shoota:
Adjusts the colors for code blocks. The style is based on [mkdocs-material](https://github.com/squidfunk/mkdocs-material/blob/02bf848fd504b953acb1bdb8aad9531ac052c93a/src/templates/assets/stylesheets/main/extensions/pymdownx/_highlight.scss) which we already use in https://github.com/crystal-lang/crystal-book/

There is still room for improvement, but this is certainly a step forward from the previous color scheme.
Some details are limited by the syntax highlighter.

Before:
grafik

After:
grafik

Dark mode on home page:
grafik


_sass/pages/_home.scss
```diff
@@ -17,6 +17,17 @@
   --light-text: hsl(0, 0%, 80%);
   --lighter-text: var(--white);
 
+  .highlight {
+    --md-code-fg-color: var(--light-text);
```
beta-ziliani, line 21: I was expecting a sort of switch/include (light/dark mode) and not a series of settings (also, why the commented lines?)

straight-shoota: It's the easiest path. Dark mode is currently only used on the homepage, so I've focused the changes there. I was thinking of incorporating a general dark mode feature, at which point this part would be refactored.

Comments are leftovers from refactoring. I'll remove them.

PR #635, opened 2024-02-07
Increase contrast for code highlighting

straight-shoota:
Some color fine tuning to increase contrast for matching the WCAG AA guideline.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast?utm_source=devtools&utm_medium=a11y-panel-checks-color-contrast

The difference is barely visible, especially in the light theme. Colors are a tot brighter in the dark theme.

Before:
grafik

After:
grafik

Before:
grafik

After:
grafik

Before:
grafik

After:
grafik

Before:
grafik

After:
grafik


beta-ziliani (approved):
<3

PR #636, opened 2024-02-07
Improve HTML semantics

straight-shoota:
There should be no visual differences from these changes.


_includes/site/footer.html
```diff
@@ -61,19 +61,19 @@ <h3>News feeds</h3>
 
     <div class="link-item">
       <a class="link-item__main" href="/feed.xml" rel="alternate" type="application/atom+xml">
-        {%- include icons/rss.svg %}
+        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><use href="#icon-rss"></use></svg>
```
beta-ziliani, line 64: I fail to see the improvement here 🤔

beta-ziliani (approved):

_includes/site/footer.html
straight-shoota: Less duplication in the rendered HTML. Which included a duplicate `id`.
I suppose we could also remove the `<use>` references alltogether and always use an include without `id` on the `<svg>` element.

PR #638, opened 2024-02-07
Use webp exclusively for event images

straight-shoota:
The png images are missing. And webp is so broadly supported that we probably don't need them.

### DIFF
--- a/_pages/community.html
+++ b/_pages/community.html
@@ -257,10 +257,7 @@ section: community
     {%- assign events = site.events | reverse %}
     {%- for event in events %}
       <article class="event-card">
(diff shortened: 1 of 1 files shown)


beta-ziliani (approved):
We should normalize or fix the viewport of the images so everyone has the same h and w.

straight-shoota:
The images should all clip to the same aspect ratio on the community page. Is that not the case for you?

beta-ziliani:
No, it's not :-)

straight-shoota:
It worked but was broken. Should be fixed by #647

PR #644, opened 2024-02-08
Order sample file names by weight

straight-shoota:
This patch puts an intrinsic ordering in the file names of the examples (instead of a `weight` attribute). This makes the ordering easily discoverable in the file tree.


beta-ziliani (commented):
Nitpick, but wouldn't `-` make more sense than `_`?

straight-shoota:
IMO one is as good as the other... What do you think makes `-` better?

beta-ziliani:
I was under the impression that most files used `-`, but no

beta-ziliani (approved):

PR #646, opened 2024-02-08
Add link to release notes on blog index

straight-shoota:
grafik


beta-ziliani:
I find it too small. The original design is better in this regard.

image

straight-shoota:
Updated:

grafik

beta-ziliani:
Is there a reason it's not in the same level as the first post, as in the design?

straight-shoota:
Yeah, positioning it there would be quite hard to implement as it needs lots of special case layouting. So I did the simple thing first. That's not ideal, of course, and we can invest into enhancing this. But we'll also need to define the layout for small and medium viewports where there's not as much space available. The current placement works well enough in all sizes, so that's the easy go.

beta-ziliani (approved):
🚀

PR #650, opened 2024-02-08
Improve examples

straight-shoota:
Improves the code examples and their descriptions.
The the examples should be small, self-contained and show a significant feature.

### DIFF
--- a/_samples/1_server.md
+++ b/_samples/1_server.md
@@ -1,8 +1,9 @@
 ---
+short_name: HTTP Server
 title: Batteries included
 description: |
-  Crystal's API includes a whole range of classes to let you start working on your project with minimal dependencies.
-read_more: '[Check the API](https://crystal-lang.org/api/)'
+  Crystal's standard library comes with a whole range of libraries that let you start working on your project right away.
+read_more: '[Check the API docs](https://crystal-lang.org/api/)'
 playground: false
 ---
 ```crystal
@@ -14,8 +15,9 @@ server = HTTP::Server.new do |context|
   context.response.print "Hello world, got #{context.request.path}!"
 end
 
-puts "Listening on http://127.0.0.1:8080"
-server.listen(8080)
+address = server.bind(8080)
+puts "Listening on http://#{address}"
 
-# (This example can't be run from carc.in)
+# This call block until the process is terminated
+server.listen
 ```
--- a/_samples/5_concurrency.md
+++ b/_samples/5_concurrency.md
@@ -2,24 +2,22 @@
 title: Concurrency Model
 short_name: Concurrency
 description: |
-  Crystal uses green threads, called fibers, to achieve concurrency. Fibers communicate with each other using channels, as in Go or Clojure, without having to turn to shared memory or locks.
-read_more: "[Read more about Crystal's concurrency model](https://crystal-lang.org/reference/guides/concurrency.html)"
+  Crystal uses green threads, called fibers, to achieve concurrency.
+  Fibers communicate with each other via channels without having to turn to shared memory or locks ([CSP](https://www.wikiwand.com/en/Communicating_sequential_processes)).
+read_more: "[Read more about concurrency](https://crystal-lang.org/reference/guides/concurrency.html)"
 ---
 ```crystal
 channel = Channel(Int32).new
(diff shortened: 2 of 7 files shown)


_samples/5_concurrency.md
```diff
@@ -2,24 +2,22 @@
...
 channel = Channel(Int32).new
-total_lines = 0
 
-sites = ["http://www.example.com", "http://info.cern.ch/"]
-
-sites.each do |site|
+3.times do |i|
   spawn do
-    channel.send site.size
+    3.times do |j|
```
beta-ziliani, line 14: Originally, the idea was to have something that really used concurrency, that's why it fetched websites (before it was fetching the sites). It'd be nice if we can think of an example in that spirit.

_samples/6_bindings.md
```diff
@@ -1,35 +1,24 @@
...
-
-  def >(other : BigInt)
-    LibGMP.cmp(pointerof(@mpz), pointerof(other.@mpz)) > 0
-  end
-end
+# Call a C function like a Crystal method:
+puts LibM.pow(2.0, 4.0) # => 16.0
 
-puts "10*100 > 20*50 😂" if BigInt.new("10"*100) > BigInt.new("20"*50)
+# This example intentionally uses a simple stdandard C function to be succinct.
```
beta-ziliani, line 21:
```suggestion
# This example intentionally uses a simple standard C function to be succinct.
```

beta-ziliani (commented):
Some examples overflow the area. I think we can add scrolls in those (I wouldn't expand the area). Maybe we can reduce a bit the top margin and reduce the height of the code to have the first example fit more gracefully.

Comments on stuff that is not touched by this PR:
- the `ENV` part in typing is a new thing that adds nothing, if we can find a more compeling yet simpler example that'd be optimal.
- The syntax example can be simplified. I don't think we need to show the `? : ` operator.

```crystal
def longest_repetition(string)
  string
    .chars
    .chunk(&.itself)
    .map(&.last)
    .max_by(&.size)
end

# press ▶️ and check the result
puts longest_repetition("aaabb")
```

straight-shoota:
> Some examples overflow the area. I think we can add scrolls in those (I wouldn't expand the area). Maybe we can reduce a bit the top margin and reduce the height of the code to have the first example fit more gracefully.

That's entirely intentional in the design to break up the layout a little bit. This is unrelated to the content edit and already the case before this change.

_samples/1_server.md
```diff
@@ -14,8 +15,9 @@ server = HTTP::Server.new do |context|
   context.response.print "Hello world, got #{context.request.path}!"
 end
 
-puts "Listening on http://127.0.0.1:8080"
-server.listen(8080)
+address = server.bind(8080)
+puts "Listening on http://#{address}"
 
-# (This example can't be run from carc.in)
```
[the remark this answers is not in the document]
straight-shoota: I'm not sure about the playground at this point. There are a couple of issues with it.
Anyway, it's not implemented yet. It might be added or not. That doesn't matter though. If we add playground integration, there simply won't be a run button for examples that don't work.

This comment doesn't doesn't help much either way and has a chance of confusing readers because it's not clear what it means. I think it's better to drop it.

_samples/5_concurrency.md
straight-shoota: Yeah, that would be great. Fetching random HTTP requests isn't very self-contained though.

Perhaps we could consider both an HTTP server and client in one program? That would be another HTTP example though. And the HTTP overhead shoudn't be necessary to showcase concurrency.

beta-ziliani: We can leave it here and change it later if we find a better example. I agree the HTTP one might be too much. Maybe one reading files?

_samples/1_server.md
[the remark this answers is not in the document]
beta-ziliani: fair enough

beta-ziliani (approved):
I'm not convinced of the visuals in the first tab, but I let that to the design team's critique. It's a small thing.

straight-shoota:
Yeah, we need to fine-tune that. The overlap visuals are interdepent on the length of all other code examples.

PR #651, opened 2024-02-08
Add page *Newsletter*

straight-shoota:
Adds a full page for the newsletter, instead of being a redirect to a fragment link on the page footer.

grafik


beta-ziliani (approved):
🚀

PR #656, opened 2024-02-09
Add page *Success Stories*

straight-shoota:
Imports existing success stories from https://github.com/crystal-lang/crystal-website/tree/relaunch-2022/_posts/success_stories
and categorize a couple more existing blog posts the same way.

grafik

The top three success stories on the home page:

grafik

The same is shown on *Used in production*.

The success story pages are normal post pages. But they have a new header image now:

grafik

The new testimonial profile component:

grafik


straight-shoota:
> It'd be nice if the success stories could be linked together instead of being linked to the rest of the posts. How hard is that to do?

I have already plans for that. It'll be a follow up.

beta-ziliani:
I don't understand the change image -> header_image :think

straight-shoota:
`image` is set to a placeholder default attribute for every page. I'll have to dig for the reason for that, but I believe it has something to do with the seo plugin.
We only want to show an image here when one is actually defined on the post.

straight-shoota:
It's probably okay to drop the `image` default property.

beta-ziliani (approved):

beta-ziliani:
one more thing: the success stories are not tied to anything else. I would assume it should be in the same navigation menu as used in prod.

straight-shoota:
Success stories are part of the blog. This will make more sense with #659 with more post categories.

PR #667, opened 2024-02-26
Refactor example tabs interaction

straight-shoota:
Refactor the example tabs implementation with a horizontal scroll pane and JS interaction.
The vertical scroll position now does not jump when selecting a tab.
Also horizontal scrolling is possible in the tab pane, allowing for an alternative navigation.

Inspired by https://web.dev/articles/building/a-tabs-component

### DIFF
--- a/_includes/pages/home/examples.html
+++ b/_includes/pages/home/examples.html
@@ -1,28 +1,28 @@
 <div class="example-tabs">
 
-  <ul class="controls">
+  <nav class="example-tabs__controls">
     {%- for sample in site.samples %}
-    <li>
       <a href="#{{ sample.slug | slice: 2, 100 }}">{{ sample.short_name | default: sample.title }}</a>
-    </li>
     {%- endfor %}
-  </ul>
+  </nav>
 
-  {%- for sample in site.samples %}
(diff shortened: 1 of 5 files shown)


beta-ziliani:
I'm completely failing to see a difference

beta-ziliani:
In FF there's no real difference, is there? And from what I can understand in Chrome the change without the js is good enough to avoid it from jumping at each click of the tab. So is it really necessary?

straight-shoota:
Yes, there's a difference and it's necessary. The JS handler prevents absolutely any vertical scrolling. This could happen even in Firefox in some situations (the viewport would jump when clicking the link while scrolled down, for example).
And horizontal scroll is pretty nice to swipe through the list, especially on touch devices.

beta-ziliani:
There's a bug when scrolling horizontally in the examples panel: the dashed border of the selected tab stays:
image


We can probably ditch the dash rounded box entirely.

straight-shoota:
That's a missing blur action because we prevent the event default. Fixed in latest commit.

beta-ziliani (approved):
🚀

PR #672, opened 2024-02-29
Improve HTML metadata and SEO

straight-shoota:


_config.yml
```diff
@@ -68,6 +68,12 @@ defaults:
 twitter:
   username: CrystalLanguage
 
+social:
```
beta-ziliani, line 71: What's the point of this? It isn't being used, right?

straight-shoota: This is configuration for the [seo plugin](https://github.com/jekyll/jekyll-seo-tag)

beta-ziliani (approved):

PR #673, opened 2024-02-29
Remove Nikola from top-sponsors and improve Manas SVG

straight-shoota:


beta-ziliani:
I see. When clicking and then swiping the dashed border stays in the clicked button:
image

Perhaps we can ditch the dash border entirely.

beta-ziliani (approved):

PR #676, opened 2024-03-01
Add component for partner images to illustrate partnerships

straight-shoota:
grafik

The rendered images are used in metadata, which powers embed previews, for example. They're just screenshots of the component which we'll have to add explicitly for each new partnership post.
We could consider automating that, but it's probably not worth it.


beta-ziliani (approved):
It's OK, although we should improve Bright's image, because it has white on the sides, making it not aligned with the distance to the + from the Manas logo.

Dark version:
image

Light version:

PR #683, opened 2024-03-05
Improve and fix text on install pages

### LINKED ISSUE #681: [Relaunch] Install page
- [ ] In System it would be nice to clarify that the version is based on the latest. People might get surprised if running the command brings an older version.
- [ ] We do keep Homebrew up to date. `Community` isn't the right place for it.
- [ ] Unrelated to the relaunch, but IIUC the FreeBSD instructions are asking to install LLVM 6.0 😮 
- [ ] Unrelated to the relaunch, but the OpenBSD instructions are asking to install PCRE1

BTW, I'll never grow tired of saying how nice this new page is!

straight-shoota:
Resolve #681

### DIFF
--- a/_pages/install/on_openbsd.md
+++ b/_pages/install/on_openbsd.md
@@ -26,8 +26,8 @@ cd /usr/ports/lang/crystal
 doas make clean install
 ```
 
-To avoid building dependencies from source (which can take a long time), you can first install them from binary packages:
+To avoid building LLVM from source (which can take a long time), you can first install the binary package:
 
 ```bash
-doas pkg_add llvm libiconv boehm-gc libevent2 pcre libyaml
+doas pkg_add llvm
 ```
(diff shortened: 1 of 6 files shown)


_pages/install/on_openbsd.md
```diff
@@ -26,8 +26,8 @@ cd /usr/ports/lang/crystal
 doas make clean install
 ```
 
-To avoid building dependencies from source (which can take a long time), you can first install them from binary packages:
+To avoid building LLVM from source (which can take a long time), you can first install the binary package:
 
 ```bash
-doas pkg_add llvm libiconv boehm-gc libevent2 pcre libyaml
```
beta-ziliani, line 32: why the other dependencies aren't needed?

straight-shoota: The packages will still be installed as they're dependencies of the port. The recommendation to install the binary package is mostly important for `llvm` which takes very long to build from source. The others are not as critical, so I don't think we need to explicitly mention them. This is the same in the FreeBSD instructions.

beta-ziliani (approved):

PR #701, opened 2024-03-06
Make `thead` sticky

straight-shoota:
This makes the table header stick to the top of the viewport when scrolling down the body.
grafik


beta-ziliani (approved):
nice catch

PR #706, opened 2024-03-07
Normalize `font-size` values

straight-shoota:
This is an internal CSS refactor with no visual changes.


ysbaddaden (approved):
**aside**: I wonder if SASS variables still make sense now that we have CSS variables (that are already used).

straight-shoota:
I don't recall exactly but there was some reason to keep the font sizes as SCSS variables. I believe some of them were used for calculations somewhere (not sure if that's still the case, though).

PR #708, opened 2024-03-07
Add taxonomy links

straight-shoota:
Adds links to the categories of each post.

grafik
grafik


beta-ziliani:
the test is failing, but it works, so I'm confused

beta-ziliani (approved):
This is a great idea 🚀

PR #718, opened 2024-03-08
Add note about package source in Repology version info

straight-shoota:


beta-ziliani:
Mmm... the information is OK, but it does not convey the message in a quick look. I was thinking in something like

straight-shoota:
@beta-ziliani That image doesn't load.

beta-ziliani:
d'oh! it was nothing fancy, just text below the System label:

image

straight-shoota:
Hm, I don't know. Seems like a very crude simplification to me. It doesn't directly depend on the OS version, but the package sources you are subscribed to.
And that limitation applies for all install sources, really. The most recent version available in any of the repositories might not be available on your specific system. It always depends on the configuration.

straight-shoota:
I'm happy to talk about solutions, but I don't see many viable ones.

The location below the group title is not good. It's visually distance from the part it refers to (the version badge). And it's not specific to any particular group. It effectively applies to *all* version badges (with varying degrees of version variability, of course).

We could put some kind of asterisk on each version badge to make it more visible. But I don't think that's a good solution either.

Maybe a general note in the introduction could do the job?

However, I think the changes in this PR should be fine and there's no dire need for more explicit clarification.
We could leave it like this for now and wait to see how users react. If there's confusion, we can still apply changes.

beta-ziliani:
> Maybe a general note in the introduction could do the job?

Yes, totally

> However, I think the changes in this PR should be fine and there's no dire need for more explicit clarification.

I disagree with the second part, but it's true that it's orthogonal to this PR

beta-ziliani (approved):

PR #719, opened 2024-03-08
Increase contrast for headers in dark mode

straight-shoota:


beta-ziliani:
I find weird that now H1(?) stand out less than H2(?)
image

beta-ziliani (approved):

PR #720, opened 2024-03-11
Fix rendering issues in Safari on iOS

straight-shoota:


straight-shoota:
Apparently the CSS selector for link action anchors was at fault: `.link-actions a:link` just wasn't applied correctly to matching elements.
But the a nested selectors such as `.link-actions a:link[href="/releases/"]` were. This made the icons show up, but they didn't get the mask image set.

I noticed some funky stuff on desktop browsers as well, where the visuals were as intended, but not all of the applied rules showed up in the dev tools. So I simplified the selector to `.link-actions > a` (which is equally correct and more efficient anyway) and now it seems to work.

The `-webkit` prefixes should also be useful for older browser versions.

This patch fixes another bug with external link icons where `display: inline` caused the background to be visible in the margins above and below the mask image in Safari on iOS.

ysbaddaden:
**suggestion** (non blocking) what about adding https://autoprefixer.github.io/ to the build?

straight-shoota:
Could do that. But I don't think it's that useful. Autoprefixing helps extend support for quite old browser versions. We should already be pretty far in the support range, despite using some relatively modern CSS features. But some of them won't be available in really old browsers at all and cannot be prefixed anyway. 
So it's probably fine to use explicit prefixes on some select properties.

PR #740, opened 2024-03-21
Refactor sponsor data and introduce small logo sponsors on homepage

straight-shoota:
grafik

Original preview:
grafik


scripts/merge.cr
```diff
@@ -65,6 +65,8 @@ end
 all_sponsors.sort_by! { |s| {-s.last_payment, -s.all_time, s.since, s.name} }
 
 write_csv("sponsors.csv", all_sponsors)
+write_csv("sponsor_logos_l.csv", all_sponsors.select { |sponsor| sponsor.last_payment.to_i > 1000 })
```
beta-ziliani, line 68: This (might) need adjustment once we get the new levels.

beta-ziliani (commented):
This is good. We need to ensure the proper level, and get a confirmation with @jkicillof that the design on the home page is good with the new tier.

jkicillof:
I don't get the new feature, there're already sponsor logos at home page design, are you planing to add another hierarchy?
image

jkicillof:
I didn't see the small logo in the preview 🤦‍♂️. I think we shouldn't have different logo sizes there. If we must do that I think we could add a badge to the best sponsors

straight-shoota:
Yeah we want to differentiate the sponsor logos based on their contribution value. A single class of sponsor logo throws everything in one basket and we want more discrete options to offer better perks for higher sponsorship tiers. This is part of an ongoing revamp of sponsorship tiers.

Differentiating on the basis of logo size is a common means to represent sponsorsship tiers. Size increments communicate a clear hiearchical order intuitively. I don't think a badge can do that efficiently.

ysbaddaden:
It's just the "works on arm" logo ain't the best logo for such a tiny size, unlike 84codes, Manas or Nikola ones, that would still fare well.

straight-shoota:
Yeah, perhaps we should increase the image size a bit to make it work better. We could consider skipping the text label then.

jkicillof:
I updated the sponsors section [here](https://projects.invisionapp.com/d/main#/console/20582111/432474609/preview). The logos on the second tier are smaller but not so much and they have less contrast.

straight-shoota:
I implemented these visuals keeping the same format but smaller logo size and lighter colors.

Getting the colors is a bit tricky, I'm using contrast and grayscale filters. It looks fine for the current set of logos.

Updated screenshot in the OP.

scripts/merge.cr
```diff
@@ -64,11 +64,15 @@ end
 
 all_sponsors.sort_by! { |s| {-s.last_payment, -s.all_time, s.since, s.name} }
 
-File.open("#{__DIR__}/../_data/sponsors.csv", "w") do |file|
-  CSV.build(file) do |csv|
+write_csv("sponsors.csv", all_sponsors.select(&.listed?))
+write_csv("sponsor_logos_l.csv", all_sponsors.select { |sponsor| sponsor.last_payment.to_i > 1000 })
+write_csv("sponsor_logos_s.csv", all_sponsors.select { |sponsor| sponsor.last_payment.to_i.in?(350...1000) })
```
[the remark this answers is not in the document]
straight-shoota: Actually, I think we should leave the old entry threshold for at least the rest of this month to keep the logo of PlaceOS around because they have already paid for it.

```suggestion
write_csv("sponsor_logos_s.csv", all_sponsors.select { |sponsor| sponsor.last_payment.to_i.in?(250...750) })
```

_data/others.json
```diff
@@ -30,5 +30,14 @@
     "last_payment": 0,
     "all_time": 425,
     "since": "Apr 14, 2019"
+  },
+  {
+    "name": "Works on Arm",
+    "url": "https://developer.arm.com/solutions/infrastructure/works-on-arm",
+    "logo": "sponsors/woa.png",
+    "last_payment": 500,
+    "all_time": 0,
```
beta-ziliani, line 39: For posterity, we're not listing the total, so there's no point in having this updated

beta-ziliani (approved):
Once the changes are committed, this is good to go

PR #751, opened 2024-03-26
Remove Nikola logo and link from sponsors list

straight-shoota:
Nikola still showed a logo and link on the sponsor page because it was hard coded in `others.json`.


beta-ziliani (approved):
We need to see how to treat these cases. I mean, they provided a lot of funding to the project, and this is something we might want to show the world. But for now this is fine.

PR #754, opened 2024-03-27
Remove broken nightly download links for Windows

straight-shoota:
The nightly downloads for Windows currently don't work.

We should enable them once https://github.com/crystal-lang/crystal/issues/14256 is resolved.

### DIFF
--- a/_data/packages-nightly.yaml
+++ b/_data/packages-nightly.yaml
@@ -65,19 +65,19 @@
 
 ### Windows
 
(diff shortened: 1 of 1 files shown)


beta-ziliani:
To me removing the entries is the proper solution. The instructions are not saying how to install the nightly, so it really doesn't add much.

beta-ziliani (approved):

PR #756, opened 2024-03-27
Fix platform logos in dark mode

GeopJr:
In dark mode, due to the platform logos being in img elements, they do not inherit the text color.

image

This PR fixes it in a hacky(?) way, by inverting their color in dark mode

image

### DIFF
--- a/_sass/mixins/_link-with-icon.scss
+++ b/_sass/mixins/_link-with-icon.scss
@@ -31,8 +31,10 @@
     mask: var(--link-icon) center / contain no-repeat;
   }
 
-  &:not(:hover) > img {
-    filter: #{"grayscale()"};
+  @media not (prefers-color-scheme: dark) {
+    &:not(:hover) > img {
+      filter: #{"grayscale()"};
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


beta-ziliani (approved):
I let Johannes take the last word, but LGTM

GeopJr:
On a second look, it seems that it was already supposed to do that but the greyscale filter overrode it

Instead I made the greyscale filter active only on not dark mode

_sass/mixins/_link-with-icon.scss
```diff
@@ -31,8 +31,10 @@
     mask: var(--link-icon) center / contain no-repeat;
   }
 
-  &:not(:hover) > img {
-    filter: #{"grayscale()"};
+  @media not (prefers-color-scheme: dark) {
+    &:not(:hover) > img {
+      filter: #{"grayscale()"};
+    }
   }
```
straight-shoota, line 38: **suggestion:** Actually, I believe that entire rule is unnecessary. It's an unintentional leftover from a feature that didn't get implemented.
```suggestion
```

None of the icons is coloured and needs greyscale. If we needed that, it would have to apply always, not only in hover state.

GeopJr: Sounds good to me!

> If we needed that, it would have to apply always, not only in hover state.

(FWIW, it's applied when *not* hovered)

straight-shoota (approved):

PR #762, opened 2024-04-02
Fix disable inline margin as example tab scroll bar workaround

### LINKED ISSUE #759: Weird behavior with overflow y scrollbars on examples
The y scrollbars are missing on both firefox and chromium by default due to `margin-inline-start` and `calc` (guessing) - even though they are supposed to be visible. This looks like a browser bug in both cases:

Firefox - toggling `margin-inline-start` fixes it:

[Screencast from 2024-03-29 16-43-21.webm](https://github.com/crystal-lang/crystal-website/assets/18014039/93c294d1-c6ff-4335-9501-c4e4631117b1)

Chromium - opening the inspector seems to fix it:

[Screencast from 2024-03-29 16-46-28.webm](https://github.com/crystal-lang/crystal-website/assets/18014039/a1813cdd-4dc0-4800-80b7-982e724ee6c9)

GNOME Web (webkit) - while it might not be 1:1 with Safari, no amount of style toggling makes the scrollbar appear

straight-shoota:
Resolves #759

### DIFF
--- a/_sass/components/_example-tabs.scss
+++ b/_sass/components/_example-tabs.scss
@@ -78,8 +78,9 @@
         row-gap: var(--padding-sm);
       }
 
-      padding-inline-start: var(--padding-xs);
-      margin-inline-start: calc(-1 * var(--padding-xs));
+      // FIXME(#759): This is disabled as a workaround for vanishing scroll bars
+      // padding-inline-start: var(--padding-xs);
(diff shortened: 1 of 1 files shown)


beta-ziliani (approved):
I guess this is fine. I fail to notice what the commented lines were supposed to be fixing.

PR #767, opened 2024-04-05
Refactor main layout centering and margins

### LINKED ISSUE #765: Page margins are lacking in Safari 15
https://forum.crystal-lang.org/t/website-relaunch/6719/8

> In Safari 15.6.1 this page https://crystal-lang.org/install/ lacks left and right margin.

https://forum.crystal-lang.org/t/website-relaunch/6719/15

> The is only 1 property marked as unsupported in Safari 15.6.1: `container-type: inline-size;` (body element, set in `layout.scss:13`)
> 
> But the zero margin problem is due to it being set in `reset.scss:11`:
> 
> 5d394c4b5f37e8ec9cc68983c2948f9f5ceeb6fa

I think this should affect *all* pages, not just install. There's nothing special about the margins on the install page.

The zero margin on `body` is correct. The margins are supposed to be on the `main` which has a `width` restriction: `min(100cqw - 3rem, var(--container-max, 120ch))`. So it should have at least 1.5rem on either side.
But I suppose container queries don't work in Safari 15.6, so the entire calculation is discarded. The container query is itself just a workaround for an incorrect calculation viewport size with scroll bar (#716).
Ideally would find a different solution/workaround for this original issue.
In the meantime, a fallback should work as well.

Safari 17 looks fine.

/cc @vlazar

straight-shoota:
This patch reverts the workaround from #716 and instead refactors the entire centering layout and margin. It's less complicated now.

Resolves #765 (hopefully)

### DIFF
--- a/_includes/posts/more_articles.html
+++ b/_includes/posts/more_articles.html
@@ -1,6 +1,6 @@
 {% assign numposts = include.posts | size %}
 {% if numposts > 5 %}
-<hr class="full-width" />
+<hr class="full-width-rule" />
 
 <section>
   <h1>More articles</h1>
--- a/_sass/_utilities.scss
+++ b/_sass/_utilities.scss
@@ -21,12 +21,6 @@
   --link-hover-color: var(--lighter-text);
 }
 
-.full-width {
-  width: 100vw;
-  transform: translateX(-50%);
-  margin-left: 50%;
-}
-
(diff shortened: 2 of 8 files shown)


beta-ziliani (approved):
LGTM but it'd be nice if some people experiencing the issue can comment on it

PR #769, opened 2024-04-05
Fix workaround for WebKit line break in `link-actions`

### LINKED ISSUE #766: Unintended line breaks in `link-actions`
There should not be any line breaks, the items are supposed to span as wide as they need to.

| Safari | Chrome |
|-|-|
| Screenshot 2024-03-28 at 12 19 11 PM | grafik |

| Safari | Chrome |
|-|-|
| grafik | grafik |

This layout appears in Safari and Epiphany. Firefox and Chrome render it correctly.

The link underline is also missing in Safari (#768).

straight-shoota:
Resolves #766

### DIFF
--- a/_sass/components/_link-actions.scss
+++ b/_sass/components/_link-actions.scss
@@ -7,5 +7,10 @@
 
   > a {
     @include link-with-icon-themed;
+
+    // FIXME: This is a workaround for WebKit weirdly introducing arbitrary
+    // line breaks
+    // https://github.com/crystal-lang/crystal-website/issues/766
+    white-space: nowrap;
   }
 }


beta-ziliani:
image

straight-shoota:
Yes, the link in the code example description is a different element. And we can't use the same workaround here because the link labels are longer here and we need line wrapping on small viewports.
We could work around by having the link span the entire available horizontal space.

The footer logo looks like a similar issues with flex elements not stretching to the appropriate `fit-content` width. I'll add a separate workaround for that.

beta-ziliani (approved):
damn WebKit!

PR #770, opened 2024-04-05
Fix example description link layout in WebKit

straight-shoota:
Fixes https://github.com/crystal-lang/crystal-website/pull/769#issuecomment-2039622593

### DIFF
--- a/_sass/components/_example-tabs.scss
+++ b/_sass/components/_example-tabs.scss
@@ -102,7 +102,11 @@
           line-height: var(--line-height-sm);
 
           > a {
-            width: fit-content;
+            // FIXME: This rule is deactivated because it's wonky in WebKit
+            // Apparently `fit-content` isn't calculated properly and the text
+            // regularly includes a line break, even though there would be enough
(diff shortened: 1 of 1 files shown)


_sass/components/_example-tabs.scss
```diff
@@ -102,7 +102,11 @@
           line-height: var(--line-height-sm);
 
           > a {
-            width: fit-content;
+            // FIXME: This rule is deactivated because it's wonky in WebKit
```
beta-ziliani, line 105: If deactivating it still looks good, is there a point in keeping it?

beta-ziliani (approved):
OK, but I have a question

_sass/components/_example-tabs.scss
straight-shoota: Having the link stretch over the entire width isn't ideal because then it expands into the blank space where there's no text or icon. This is not necessarily *wrong* and we could just assume this as expected. But I'd rather have it fit more snugly.

grafik

Issue #758, opened 2024-03-28
New site doesn't load on Mobile Safari


jwoertink:
Love the new design :raised_hands:  but I'm having trouble getting it to load on mobile. I'm using iOS 17.4.1 on iPhone 13.

This is after clearing cache, and I've also tried opening up in private browsing.

![IMG_2582](https://github.com/crystal-lang/crystal-website/assets/2391/565e5592-dbfb-4b00-87f8-830d90210dc6)

After a few attempts, I was able to get the logo to load as well, but still nothing else.

straight-shoota:
That's odd. We had a test with iOS 17.3.1 where everything was fine. 🤔

jwoertink:
I tried connecting to the safari debugger to see if I could get anymore info, but the network panel looked fine, and there weren't any issues in the console... Though, I'm not sure I trust the debugger since updating my mac :confused:  I have issues with it on my own site. I'll see if I can get any other info on this.

straight-shoota:
I did another test with iPad Pro 12.9", 2nd Generation iPadOS 17.2 (21C62) and everything looks fine there.

jwoertink:
Yeah, it's still doing it for me. One weird thing I noticed is the site actually loads and I see all of the elements for a split second, then they all go away except for 1 or 2 parts. Then if I try to reload the page, it just never loads. No plugins or extensions, and it happens using Wifi and 5G networks. I can do a screen recording and send it if you want. Not sure if github will let me submit a video in a comment. If you'd like, let me know where I can send a screen recording.

jwoertink:
Just to make sure it's not just me, it looks like @russ gets the issue too

![image](https://github.com/crystal-lang/crystal-website/assets/2391/71f37abb-ca52-4a73-b31b-0ad50ef2630d)

However, the other people on my team seem to load the site fine. Clearing cache on the entire browser and private browse tab still doesn't fix it.

russ:
Just to note, the site does load on mobile, but just content above the fold. Scrolling down you see everything just blank after Johannes Müller. iPhone 11, iOS 17.3.1

straight-shoota:
We received another report here: https://fosstodon.org/@fnordfish@ruby.social/112205967637696757

fnordfish:
iPhone 11 Pro, iOS 17.4.1

https://ruby.social/@fnordfish/112205967220293581

jwoertink:
> reload seems to be blocked.

[...]

PR #780, opened 2024-04-15
Disable styles that break rendering on Safari iPhone

straight-shoota:
This is a quick fix to enable rendering on iPhone Safari. It degrades visuals on all platforms (the pattern doesn't span the entire widht on large viewports), but this is acceptable as a temporary fix.
We'll need to investigate if we can find other solutions or make this one selective to only affected browsers.

See https://github.com/crystal-lang/crystal-website/issues/758#issuecomment-2053688906


beta-ziliani (approved):
Bloody Safari. Thanks Jeremy!

PR #785, opened 2024-05-06
Rename title for Standard Library section

Fryguy:
In the new layout it is strange to me that the very first thing you see in the list of features is "HTTP Server".  It makes it sound like our most important feature is an HTTP server, and I feel like it gives a wrong first impression. Instead, I think we want to convey that we have a strong standard library, and then we can use the http server as the example internally.

So, this PR just renames the title to "Standard Library".

I was going to open an issue for discussion, but the change was so small, I figured we can discuss here and close if we don't like it. Separately, I do wonder if the stdlib is really the first thing we want to present. To me, Typing, Nil Checking, and Syntax would come before it in importance (with Syntax probably first in my mind), but we can discuss that in a separate issue.

Before:

The_Crystal_Programming_Language

After:
The_Crystal_Programming_Language


beta-ziliani:
I agree to most of what you say, @Fryguy , thanks for bringing this. For this particular example, I think "Rich stdlib" might be a better name?

Fryguy:
Yup, I can do that.

Fryguy:
Done - I also update the OP screenshot and text.

straight-shoota (approved):

beta-ziliani (approved):

PR #791, opened 2024-05-20
Fixing base url for feed content

beta-ziliani:


straight-shoota:
What's the reason for this change?
I'd expect the base URL to be the path to the post, not the home page, to make path-relative links work.

beta-ziliani:
but what is `/2024/04/09/1.12.0-released/` in this context?

straight-shoota:
It's a fully qualified URL (note the `absolute_url` filter): `https://crystal-lang.org/2024/04/11/1.12.1-released/`
So any relative link inside the content would be properly scoped to this context, not the root of the website.

PR #795, opened 2024-05-31
Fix `sponsors_logos_l` to require a logo

straight-shoota:
Ref https://github.com/crystal-lang/crystal-website/pull/792/files#r1622646393

PR #797, opened 2024-06-04
Adding empty image for sponsor with no logo

beta-ziliani:
Replaces #795


beta-ziliani:
Done that, although the merge system is not great for this. We should implement a better strategy in the future; one that doesn't require so many fields when just needing to override one.

straight-shoota (approved):

straight-shoota:
@beta-ziliani I understand partial merge should work with `overrides` property... I believe it did in the past. But might need to look into. We can merge this of course.

PR #852, opened 2024-10-08
[CI] Change `update-sponsors` workflow to run weekly every Tuesday

straight-shoota:
I think it's better to have the workflow run every week than on fixed days of the month which leads to an irregular schedule. I'm chosing Tuesday because today is a Tuesday. And it seems convenient.

### DIFF
--- a/.github/workflows/update-sponsors.yml
+++ b/.github/workflows/update-sponsors.yml
@@ -2,7 +2,7 @@ name: Update Sponsors
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


straight-shoota:
Ah no, this won't do because our sponsors update script expects to be running on the first day of the month.

ysbaddaden (approved):

straight-shoota:
This is ready to merge after #854 which removes the expectation to run on the first day of the month.

PR #869, opened 2024-11-14
Add post: Lifetime Event Loop

straight-shoota:
This is a draft for announcing the event loop updates once https://github.com/crystal-lang/crystal/pull/14996 is merged.

### DIFF
--- a/_posts/2024-11-05-lifetime-event-loop.md
+++ b/_posts/2024-11-05-lifetime-event-loop.md
@@ -1,0 +1,89 @@
+---
+title: "Lifetime Event Loop"
+author: straight-shoota,ysbaddaden
+summary: ""
+categories: technical
+tags: [feature, eventloop, concurrency]
+---
+
+A core component of Crystal's concurrency model is the **event loop**. It
+integrates asynchronous operations into the runtime, enabling other fibers to
+run while one fiber waits to read data on a socket, for example.
+
+We're changing how the event loop operates under the hood on Unix systems in [#14996].
+
+The new implementation integrates the Crystal event loop directly with system
+selectors, [`epoll`](https://linux.die.net/man/7/epoll) (Linux, Android) and
+[`kqueue`](https://man.freebsd.org/cgi/man.cgi?kqueue) (BSDs, macOS) instead of
+going through [`libevent`](https://libevent.org/).
+
+In the process we're changing that file descriptors now stay subscribed to the
+event loop for their entire lifetime, instead of being added and removed on
+every single operation. This reduces overhead and improves performance.
+
+This post highlights the relevant information for users.
+
+More technical details are available in [RFC #0009].
+
+## Effects
+
+The new implementation has been merged into `master` and is available in
+[nightly builds](/install/nightly).
+
+No changes in user code are required, everything plugs right in.
+
+The new implementation is supported on Linux, macOS, FreeBSD and Android and
+automatically enabled on these systems. Read [more about
+availability][availability] in the RFC.
+
+All other operating systems keep the previous event loop driver: `libevent` on
+Unix systems, `IOCP` on Windows.
+
+In case of performance regression or other issues, you can switch
+back to the legacy event loop implementation with the compile-time flag
+`-Deventloop=libevent`.
+
+Dropping `libevent` removes an external runtime dependency from Crystal
+programs.
+
+## Caveats
+
+In some cases the new implementation may cause issues. If you're affected,
+we recommend falling back to `libevent` with `-Deventloop=libevent`.
+
+We are aware of some scenarios but believe they are quite rare and should not
+hinder general availability of this new feature. Exposure through nightly builds
+should help us gather more usage data to assess whether there are any more
+noteworthy implications we have not been aware of.
+
+### Timers and Timeouts
+
+At this point, there's a missing optimization for timers (`sleep 1.second`) and
+timeouts (`socket.read_timeout = 1.second`). In case you're using lots of them,
+it might lead to performance degradation.
+
+This is only a temporary limitation until we finish the implementation of an
+efficient data structure. We're already working on that.
+
+### Multi-Threading
+
+The new implementation works well with the multi-threading preview
+(`-Dpreview_mt`), having one event loop instance per thread.
+
+There is on caveat though: Moving file descriptors with pending operations
+between event loop instances – i.e. between threads – is an error.
+
+For example: Let's assume fiber `A` running on thread `X` waits on file
+descriptor `4`. While this operation is still pending, fiber `B` running on
+thread `Y` tries to start another operation on file descriptor `4`. This is now
+going to raise. File descriptors can only be owned by a single event loop
+instance.
+
+This limitation will be mitigated with the arrival of execution contexts from
+[RFC #0002] which share one event loop instance between all threads in a
+context.
+
+[#14996]: https://github.com/crystal-lang/crystal/pull/14996
+[availability]: https://github.com/crystal-lang/rfcs/blob/rfc/lifetime-event_loop/text/0009-lifetime-event_loop.md#availability
+[RFC #0002]: https://github.com/crystal-lang/rfcs/pull/2
+[RFC #0009]: https://github.com/crystal-lang/rfcs/pull/9
--- a/_posts/2024-11-05-lifetime-event-loop.md
+++ b/_posts/2024-11-05-lifetime-event-loop.md
@@ -1,9 +1,12 @@
 ---
-title: "Lifetime Event Loop"
+title: "A new Event Loop for UNIX operating systems"
 author: straight-shoota,ysbaddaden
-summary: ""
+summary: >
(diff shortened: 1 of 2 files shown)


_posts/2024-11-05-lifetime-event-loop.md
```diff
@@ -0,0 +1,89 @@
...
+[nightly builds](/install/nightly).
+
+No changes in user code are required, everything plugs right in.
+
+The new implementation is supported on Linux, macOS, FreeBSD and Android and
+automatically enabled on these systems. Read [more about
+availability][availability] in the RFC.
+
+All other operating systems keep the previous event loop driver: `libevent` on
+Unix systems, `IOCP` on Windows.
```
ysbaddaden, line 40:
```suggestion
For the time being other UNIX operating systems still use the `libevent` event loop by default (regressions, issues or untested). You can still force enable it using `-Devloop=kqueue` (BSDs) or `-Devloop=epoll` (e.g. Solaris).

Windows is unaffected and keeps using `IOCP`.
```

_posts/2024-11-05-lifetime-event-loop.md
```diff
@@ -0,0 +1,89 @@
...
+
+In case of performance regression or other issues, you can switch
+back to the legacy event loop implementation with the compile-time flag
+`-Deventloop=libevent`.
```
ysbaddaden, line 44: Maybe merge this paragraph with the caveat paragraph below?

_posts/2024-11-05-lifetime-event-loop.md
```diff
@@ -0,0 +1,89 @@
...
+We're changing how the event loop operates under the hood on Unix systems in [#14996].
+
+The new implementation integrates the Crystal event loop directly with system
+selectors, [`epoll`](https://linux.die.net/man/7/epoll) (Linux, Android) and
+[`kqueue`](https://man.freebsd.org/cgi/man.cgi?kqueue) (BSDs, macOS) instead of
+going through [`libevent`](https://libevent.org/).
+
+In the process we're changing that file descriptors now stay subscribed to the
+event loop for their entire lifetime, instead of being added and removed on
+every single operation. This reduces overhead and improves performance.
```
ysbaddaden, line 22: Maybe something like:

> Not short of removing an external dependency, and taking control over a core feature, this also changes how file descriptors are treated: instead of being added and removed on every blocking IO operation, the file descriptors are now added once and kept for their full lifetime, which is how `epoll` and `kqueue` have been designed.

_posts/2024-11-05-lifetime-event-loop.md
```diff
@@ -0,0 +1,89 @@
...
+
+Dropping `libevent` removes an external runtime dependency from Crystal
+programs.
+
+## Caveats
+
+In some cases the new implementation may cause issues. If you're affected,
+we recommend falling back to `libevent` with `-Deventloop=libevent`.
+
+We are aware of some scenarios but believe they are quite rare and should not
```
ysbaddaden, line 54: Maybe you meant "potential regressions" or "issues" instead of "scenarios"?

_posts/2024-11-05-lifetime-event-loop.md
```diff
@@ -0,0 +1,89 @@
...
+(`-Dpreview_mt`), having one event loop instance per thread.
+
+There is on caveat though: Moving file descriptors with pending operations
+between event loop instances – i.e. between threads – is an error.
+
+For example: Let's assume fiber `A` running on thread `X` waits on file
+descriptor `4`. While this operation is still pending, fiber `B` running on
+thread `Y` tries to start another operation on file descriptor `4`. This is now
+going to raise. File descriptors can only be owned by a single event loop
+instance.
```
ysbaddaden, line 80: Maybe outline:

> otherwise we can still "move" or "transfer" a socket or file across fibers without risks.

ysbaddaden (commented):
Nice! I left a few suggestions, plus:

- maybe change the title to "A new Event Loop for UNIX operating systems", so we don't give the impression that we completely overhauled the evloop;
- maybe add a note about `io_uring` (not forgotten, we have plans).

Last but not least: the flag is currently `-Devloop` not `-Deventloop`. I don't have strong feelings about either, but we need to settle on one.

_posts/2024-11-05-lifetime-event-loop.md
```diff
@@ -0,0 +1,103 @@
...
+
+This is only a temporary limitation until we finish the implementation of an
+efficient data structure. We're already working on that.
+
+### Multi-Threading
+
+The new implementation works well with the multi-threading preview
+(`-Dpreview_mt`), having one event loop instance per thread.
+
+There is on caveat though: Moving file descriptors with pending operations
```
beta-ziliani, line 78: As with the RFC, the breaking change should be shouted out. Also, the distinction between transferring FDs without pending operations should stand. More like "It is possible to transfer a file descriptor between threads **as long as there are no pending operations**."

beta-ziliani: This last bit is to not make it sound like trasferring is not possible at all (for the casual reader)

zw963:
Cool, if we removes the `libevent` as a runtime dependency, does it mean we can statically compile a binary on Linux use GNU toolchain now?

zw963:
> Last but not least: the flag is currently -Devloop not -Deventloop

I consider latter is better?

straight-shoota:
> Cool, if we removes the `libevent` as a runtime dependency, does it mean we can statically compile a binary on Linux use GNU toolchain now?

No, this doesn't affect that. The trouble with static linking is with the GNU libc itself. Linking `libevent` statically is no issue.

straight-shoota:
> But, Why it still depends on libevent?

The PR mentioned in this draft post has not been merged yet, so the change is not yet available in master/nightly.
It'll be when the post is published. Until then you need to use the PR branch.

_posts/2024-11-05-lifetime-event-loop.md
beta-ziliani: @straight-shoota have you seen this?

_posts/2024-11-05-lifetime-event-loop.md
```diff
@@ -0,0 +1,103 @@
...
+For example: Let's assume fiber `A` running on thread `X` waits on file
+descriptor `4`. While this operation is still pending, fiber `B` running on
+thread `Y` tries to start another operation on file descriptor `4`. This is now
+going to raise. File descriptors can only be owned by a single event loop
+instance _at a time_. We can still transfer sockets and files without pending
+operations across fibers without risks.
+
+This limitation will be mitigated with the arrival of execution contexts from
+[RFC #0002] which share one event loop instance between all threads in a
+context.
```
ysbaddaden, line 90: Indeed, the first time I read the phrase I missed the `pending` word :sweat_smile: 

Maybe:

```suggestion
There is one caveat though: file descriptors can only be owned by a single
event loop instance _at a time_. A file descriptor _can be moved_ from one
fiber to another, possibly moving from one event loop instance to another
along the way (transparently), but this is only possible if there are no
pending operations on that file descriptor.

For example: Let's assume fiber `A` running on thread `X` waits on file
descriptor `4`. While this operation is still pending, fiber `B` running on
thread `Y` tries to start another operation on file descriptor `4`. This is now
going to raise.

This limitation will be mitigated with the arrival of execution contexts from
[RFC #0002] which share one event loop instance between all threads in a
context.
```

ysbaddaden: This is mostly the same, but reordered :)

ysbaddaden (approved):

_posts/2024-11-05-lifetime-event-loop.md
straight-shoota: The suggested diff looks odd because it seems to strip off the last line of the preceding paragraph. Is that intended?

ysbaddaden: Sigh, GitHub seems to have cut the first line when selecting :face_exhaling: 

Not intended, indeed.

_posts/2024-11-05-lifetime-event-loop.md
```diff
@@ -0,0 +1,107 @@
...
+This is only a temporary limitation until we finish the implementation of an
+efficient data structure. We're already working on that.
+
+### Multi-Threading
+
+The new implementation works well with the multi-threading preview
+(`-Dpreview_mt`), having one event loop instance per thread.
+
+There is on caveat though: Moving file descriptors with pending operations
+between event loop instances – i.e. between threads – is an error.
```
beta-ziliani, line 79: Leftover.

```suggestion
```

beta-ziliani (changes requested):
I would also move the section about MT before timers. The program changing behavior is more important than a not that much noticeable performance regression.

beta-ziliani (approved):

beta-ziliani (approved):

PR #711, opened 2024-03-07
Add tags

straight-shoota:
Blog index:

grafik
grafik

Tag page:

grafik

Post page:

grafik

PR #871, opened 2024-11-14
Add *Tags* index page

straight-shoota:
This is a follow-up to https://github.com/crystal-lang/crystal-website/pull/711 adding an index for all the tags.
Previously, tags were only shown on each post and there are individual pages for each tag showing the tagged posts. This patch adds an overview page listing _all_ tags.
This is useful even just as an internal editing tool 😛 

Tags with most articles are ordered on top.
The most recent three articles per tag are rendered, and a link to the tag page.

grafik


_pages/blog/tags.md
```diff
@@ -0,0 +1,30 @@
+---
+title: Tags
+layout: page
+section: tags
+---
+
+{% capture counts_with_tags_string %}{% for tag in site.tags %}{{ tag[1] | size | prepend:"000000" | slice:-6,6 }}:{{ tag[0] }}{% unless forloop.last %},{% endunless %}{% endfor %}{% endcapture %}
```
beta-ziliani, line 7: really there isn't a better way to format a number?? 😮

beta-ziliani (approved):
Nice!

_pages/blog/tags.md
straight-shoota: Nope. Apparently not 🤷

ysbaddaden: We probably can, but that would need to write a template helper function.

straight-shoota: Yes, of course with a plugin we could make this cleaner.

PR #912, opened 2025-04-04
Add newsletter template

straight-shoota:


mverzilli:
I guess you're aware of this, but just in case, some code is spilling to the page: 

Screenshot 2025-04-08 at 12 36 59

straight-shoota:
That's intentional. The newsletter email content is intended for distribution via Mailspring and contains template expression for the Mailspring engine.

PR #940, opened 2025-05-09
Overhaul samples

### LINKED ISSUE #677: [Relaunch] Suggested changes in code snippets
## Syntax:

1. It was wrong (crashed with empty string)
2. It was suggesting to press play, but no play button so far

Suggestion:

```cr
class String
  def longest_repetition?
    max = chars
            .chunk(&.itself)
            .map(&.last)
            .max_by?(&.size)

    {max[0], max.size} if max
  end
end

puts "aaabb".longest_repetition? # => {'a', 3}
```

## Concurrency

Add some forced sleep to enforce the notion of concurrent events

Suggestion:

```cr
channel = Channel(Int32).new

3.times do |i|
  spawn do
    3.times do |j|
      sleep rand(100).milliseconds
      channel.send 10 * (i + 1) + j
    end
  end
end

9.times do
  puts channel.receive
end

### PR DESCRIPTION
**Pull request by straight-shoota**
The reasons are explained in https://github.com/crystal-lang/crystal-website/issues/677#issuecomment-2866510397

Resolves #677

### CODE REVIEW

File: _samples/1_stdlib.md [hunk 1]
```diff
@@ -7,17 +7,13 @@ read_more: '[Check the API docs](https://crystal-lang.org/api/)'
...
+require "xml"
 
-server = HTTP::Server.new do |context|
-  context.response.content_type = "text/plain"
-  context.response.print "Hello world, got #{context.request.path}!"
-end
+response = HTTP::Client.get("https://crystal-lang.org")
+html = XML.parse(response.body)
+node = html.xpath_node(%(//div[@class="latest-release-info"]//strong))
+version = node.try(&.text)
```
ysbaddaden, line 16: **Thought**: The sample is nice, but I'm not sure we should showcase XML that requires an external lib (and we're not sure is fiber safe). Could we showcase JSON instead? It's more modern and pure Crystal.

_samples/2_type_system.md
```diff
@@ -1,23 +1,17 @@
...
+short_name: Types
 title: Type system
 description: |
-  Crystal is statically typed and type errors are caught early by the compiler,
-  eliminating a range of type-related errors at runtime.
+  The compiler catches type errors early. Avoid the [billion-dollar mistake](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare).
 
-  Yet type annotations are rarely necessary, thanks to powerful type inference.
-  This keeps the code clean and feels like a dynamic language.
+  The code is still clean and feels like a dynamic language.
```
ysbaddaden, line 7: **Thought**: I wouldn't drop the info that types are inferred. It's one of the main trait of the language (c.f. Beta's talk in Sweden).

**Thought**: The billion dollar mistake link is maybe interesting but only talks to the few developers that know about it. I don't. What I know is what NIL and NULL errors are, because I experienced 'em, badly, in production.

_samples/6_bindings.md
```diff
@@ -1,8 +1,7 @@
 ---
 title: C-bindings
 description: |
-  Crystal allows to define bindings for C libraries and call into them.
-  You can easily use the vast richness of library ecosystems available.
+  Bindings for C libraries makes it easy to use existing tools.
```
ysbaddaden, line 4: **Suggestion**: specify that it's "direct C bindings", that is "without FFI or any runtime cost".

_samples/7_macros.md
```diff
@@ -4,21 +4,24 @@ description: >
...
 ---
+{% raw %}
+
 ```crystal
-class Object
-  def has_instance_var?(name) : Bool
-    {% raw %}{{ @type.instance_vars.map &.name.stringify }}{% endraw %}.includes? name
+macro upcase_getter(name)
+  def {{ name.id }}
+    @{{ name.id }}.upcase
```
ysbaddaden, line 12: **Thought**: it's simpler, but not as powerful as the current example showcases (complex macro expressions & type reflection).

_samples/1_stdlib.md
straight-shoota: Good idea. I suppose we can fetch https://crystal-lang.org/api/versions.json instead

_samples/2_type_system.md
straight-shoota: I thought we could drop inference here because it's already mentioned in the intro tab. Though we might want to edit that as well.
Do you think it's important to mention it explicitly again here? The example should _show_ this already, so we may not necessarily need to _tell_.

straight-shoota: I agree on the billion dollar mistake. It's a remnant from merging content from `3_null.md`. We should just drop it and call it null pointer exception (or something explicit like that).

_samples/7_macros.md
straight-shoota: The old example is a good showcase of macro features. But I doubt it's a good example to show the motivation for macros. It might be confusing why you would implement a custom `has_instance_var?(name)` method in the first place, when other languages (e.g. Ruby) include this directly in their reflection API.

Maybe we can find a better example for macro usage that is useful, shows some advanced features but still fits into a couple LOC.

_samples/2_type_system.md
ysbaddaden: There's nothing wrong with a dup, we may have scanned over the info in the presentation line.

_samples/6_bindings.md
ysbaddaden: Sure, what's important is the "no FFI" part.

ysbaddaden (approved):
Let's move on with this :+1:

Issue #928, opened 2025-04-24
Check external links


straight-shoota:
We previously had a CI workflow to check external links. But it was putting a big load on CI response time and was mostly ignored anyway.
So we dropped that in #921.

Link checking is notoriously difficult. It's a huge task with thousands of links, and it's quite common to see some temporary availability issues here and there.
Also, an increasing number of hosts rejects automated requests to protect against overburdoning scrapers.

This issue is to find a better solution.

We could consider moving this to a separate workflow, maybe on a schedule and not allowed to fail. Then we'd at least know when something is broken.

It could also be useful to check links already in Markdown (https://lychee.cli.rs/ can do this) and only in files that are changed in a pull request. This ensures that links are at least valid at the time of creation.

straight-shoota:
I believe we should also remove the dead links plugin (added in #637).
It automatically converts links marked as dead with a link to the web archive. But it's not fully automatic, we still need to manually mark links as dead. And it's not fail-safe because links might not exist in the archive, or contain unrelated content.
Instead, we should replace dead links directly with archive links in the sources.

HertzDevil:
Somewhat related, we might want to do the same in the main Crystal repository to catch dead links in the standard library docs, to take care of things like https://github.com/crystal-lang/crystal/issues/15800

PR #957, opened 2025-07-18
[CI] Add workflow to check new links

straight-shoota:
This is a new approach to check links.
It operates directly on the sources (i.e. without rebuilding the site) and only checks links that are newly added in a PR. This makes the check very efficient and fast, and it only cares about the current diff. This makes it much more useful than the previous approach, which we ditched in https://github.com/crystal-lang/crystal-website/pull/921.

https://lychee.cli.rs/ is also much more efficient than htmltest which we used previously.

Resolves the main part of #928.
Checking only new links does not protect against link rot, though. We might want a separate CI job for that.


straight-shoota:
I removed the action that compared links from before and after the change. Instead, I've added lychee integration for `pre-commit` which is much easier.
The semantics are a bit different because pre-commit runs on each modified file and does not ignore links that existed previously.

PR #962, opened 2025-07-22
Cleanup frontmatter fields

straight-shoota:
Removes the unused `thumbnail:` field and empty fields.


straight-shoota:
The pre_commit CI failure is unrelated to this change.

PR #984, opened 2025-09-29
Wind of change (or maybe not so)

beta-ziliani:


straight-shoota:
I pushed a commit to update the team page accordingly.

I think the team lead should be on the first position (before the Steering Council members) but we can swap things around if someone thinks it should be different.

PR #1069, opened 2026-05-17
Update pacman command to use -Syu for upgrades

ShalokShalom:
The previous command has the potential to break the entire operating system, since it takes packages that are eventually not compatible with the current (outdated) one.

Rule number one of the rolling club: 
Always update, before you install something new. 🧼

### DIFF
--- a/_data/packages.yaml
+++ b/_data/packages.yaml
@@ -52,7 +52,7 @@
   title: Pacman (Arch Linux)
   example: |
     ```shell
-    pacman -S crystal shards
+    pacman -Syu crystal shards
     ```
   repo_href: https://archlinux.org/packages/extra/x86_64/crystal/
   repo_description: Crystal package on Arch Linux package index


HertzDevil:
I assume the same holds for MSYS2's Pacman as well?

straight-shoota:
I don't understand what exactly is the problem with `pacman -S crystal`.
The documentation does not warn about that specifically. Only to avoid `-Sy` without `u` (https://wiki.archlinux.org/title/Pacman#Installing_packages).

ShalokShalom:
If you issue `-S` with `-Sy` or `-Syy` before, and no update in between, you confuse the shared libraries, due to dynamic linking. 

In case you run `-S` on a system that is not up to date, and the version of Crystal and/or Shards has changed, you will run into 404 issues. 

Both issues on its own are reason enough, to simply recommend `-Syu` - which runs always, and makes never problems.

straight-shoota:
My own Arch experience is a bit dated and I couldn't find any indication from official documentation or the Arch community that `-Syu` should be preferred over `-S`.

If we want to push this forward, we need some evidence that the change is correct.
@ShalokShalom I trust your word, but I can't tell if it's personal opinion or commonly accepted.

ShalokShalom:
I just told you, what the problem is. 

Running `-S` without prior update of the database results in a failed installation. 

Especially people who are new to Arch (CachyOS attracts many people from Windows these days) dont really know how to operate a rolling system. 

If they run `-Sy` or `-Syy` etc without updating afterwards, is a `-S` deadly. 
It not that `-S` is wrong, its simply the less safe option, if you take inexperience into account.

ysbaddaden:
I don't know about Archlinux, but I had to run `pacman -Suy` on Windows MSYS2 to get a working `crystal.exe` after upgrade.

Without `-u` it only updated the crystal package itself, and it didn't bother to update the required dependencies to working versions, and the `crystal.exe` was missing a dozen DLL files :astonished:

I assume pacman, at least the one in msys2, doesn't care about dependency versions, and assumes everything **must** be up-to-date. Why a package manager wouldn't bother with resolving dependencies is beyond me :speak_no_evil: 

I only seldom use the environment, and don't bother upgrading it until I want to upgrade one utility. I guess rolling releases want us to upgrade everyday, or to always upgrade everything. Sigh.

straight-shoota:
Okay, suggesting `-Syu` might be better for occasional users who are not familiar with the details of the package manager. Let's merge this.

We should change the pacman args for msys2 packages as well, though.

ShalokShalom:
> I don't know about Archlinux, but I had to run `pacman -Suy` on Windows MSYS2 to get a working `crystal.exe` after upgrade.
> 
> Without `-u` it only updated the crystal package itself, and it didn't bother to update the required dependencies to working versions, and the `crystal.exe` was missing a dozen DLL files 😲
> 
> I assume pacman, at least the one in msys2, doesn't care about dependency versions, and assumes everything **must** be up-to-date. Why a package manager wouldn't bother with resolving dependencies is beyond me 🙊
> 
> I only seldom use the environment, and don't bother upgrading it until I want to upgrade one utility. I guess rolling releases want us to upgrade everyday, or to always upgrade everything. Sigh.

You have a local database of the packages, and then there is the reality of the actual packages on the server.

So, now is the question, how you sync the two. The basic philosophy of Arch is, that the user has to understand why they do, what they do.

They see that as a way to both educate the user, and also enable upmost control of the system.

Personally, I think they sometimes go overboard with that, since sensible defaults should also work for children, elderly, and so on, otherwise its gatekeeping. 

And yes, rolling systems expect to be updated frequently, - you may say on a *rolling* basis. 😄 

All the packages update once a new stable version is released, and thats also the most simple distribution strategy imagineable - everything else introduces obviously friction and workload. 

Once every 2 weeks is recommended, and some people update every day, although I do I remind you, that you need to sometimes reboot the system, eg after a (major) update of systemd, or other critical systems. 🙂 

Here is a short help about your described issue: 

The `-S` stands for *synch*ronise, and all its uses synch something.

With `y` you sync your local database of packages with the database of the package server. 
The `u` flag syncs the actual packages, so to speak.

What happened in your case possibly, is that the local database had no notice yet of the updated DLL files, and adding the `y` updated that info to your local database.

Pacman treats the local database as the source of truth, so you updating via `u` updates just the ones that are available for update according to the local database. 

I wrote my own abstraction layer over that, reduces the whole complexity to a single command (`add`) and it automatically updates the system, if I need to in order to install a package. 

Arch can very much be a superb experience with grandmas and kids, its just the culture within Arch that sits inbetween that. 

Garuda provides a nice out of the box experience for desktop Arch. 🙂

ysbaddaden:
Why not always update _everything_ rather than happily installing broken packages, and leave users to guess? That's beyond elitist :face_exhaling:

ShalokShalom:
> Why not always update _everything_ rather than happily installing broken packages, and leave users to guess? That's beyond elitist 😮‍💨

That's exactly what everybody is recommending, and what I am doing as well. The tooling just allows to do stupid stuff, and I don't agree with it.

Garuda also does remind you about this, if you ever run into it, and does provide a simple, single click solution on a flashing widget, to get out of it. 

Its a bit like the whole footgun topic in programming languages, where dangerous operations are possible, and sometimes even encouraged by how easy they are to perform.

I wrote a script, to get around this:
https://codeberg.org/ShalokShalom/add-and-friends#add

PR #1082, opened 2026-06-05
Replace `markdownlint` with `rumdl`

straight-shoota:
[rumdl](https://rumdl.dev/) is a Markdown linter similar to [markdownlint](https://github.com/igorshubovych/markdownlint-cli) but with some significant improvements.
It is compatible with markdownlint, supporting the same rules and configuration as before. It's a drop-in replacement with more performance (Rust vs Javascript).
And it brings a bunch of additional rules and features, such as per-file-ignores.
See https://rumdl.dev/markdownlint-comparison/ for a full comparison.

This is a follow-up to #1080


ysbaddaden:
I've been using it locally and was gonna propose to use it :+1:

ysbaddaden (approved):

PR #1097, opened 2026-07-16
Decrease line height in code blocks

straight-shoota:
Code reads better with a smaller line height.

| before | after |
|-|-|
| image | image |


ysbaddaden:
This is debatable, but I guess this is closer to how code is rendered in editors and terminals :+1:

ysbaddaden (approved):

PR #1100, opened 2026-07-22
Enable MD060: Table format

straight-shoota:
This table format style improves readability of the markdown source code. 

We're using it in crystal-book already(https://github.com/crystal-lang/crystal-book/pull/932)

> This table format aligns all pipe characters at the same column, which makes for a tabular layout even in the source code.
It's not strictly better than the previous compact format. It certainly causes longer line lengths.
But I think the alignment improves readability. The cost for that is sometimes you need horizontal scroll. But I prefer that clarity over the tightly compressed text.

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -70,3 +70,10 @@ no-bare-urls: false
 # MD045 - Add Descriptive Text to Images
 # FIXME: Deferred
 no-alt-text: false
+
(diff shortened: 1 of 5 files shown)


ysbaddaden (approved):

ysbaddaden:
I prefer not writing into this format, but I guess the formatter can handle it.

PR #1143, opened 2026-08-04
Drop unreachable URLs from `used_in_prod.md`

straight-shoota:
When the URLs are not reachable, the project either isn't alive anymore, or changed significantly so that we cannot expect it to still use Crystal in production.

I detected these URLs based on HTTP status code. There's no guarantee that any other URL which returns a success code still contains the expected content.


manumoreira:
LGTM! 
~I wonder if it makes sense to use something like lychee to check http status periodically~ 
Just seen the GHA for lychee, let me know if I can help with that.

straight-shoota:
👉 #1147

PR #1147, opened 2026-08-04
CI: Check all links

straight-shoota:
Updates the CI tests to validate all links, not just internal ones.

I'm not sure if we want to run this on every pull request, though.

The upside is that we quickly learn about a broken link.
The downside is that such a failure is _usually_ unrelated to the PR itself (the lint workflow already checks newly added links in the source code).
And there might be some flukes due to temporary unavailability.


straight-shoota:
The remaining CI failures are due to #1143. So the link validation seems to be working 👍

ysbaddaden:
I guess there could be a weekly or monthly CRON against master, and maybe a check for new/changed links in the current PR, but yeah, not a check of every links on every PR.

